PR #34, opened 2019-04-17
Update app_fr.ts

qwertygc:
Fix translation (it's not perfect)


ccbogel:
Thank you. I will convert to the machine-readable and upload that too.

PR #39, opened 2019-09-03
make a easier maintainable project

mortbauer:
Hello Colin

I made quite big changes in that pull request because I factored out all the third party libraries to make this a bit more like a standard application.

I also added a debian directory to cover the deb building.

Would have couple of more ideas, but lets see for your opinion on the changes.

greets martin


ccbogel:
Thank you very much Martin.
I am glad you have so much interest in this project.
Before I merge the pull request. For users, who are on IOS or Windows, will some of the changes to pdfminer, ebooklib affect them? And if yes what is best for instillation instructions for them.
regards Colin

mortbauer:
Hm, it should really not cause any problems there, as all these can be installed from pypi. Great you merged it so fast! I could also give it a test on windows.

ccbogel:
Hi again, yes I can partly test on windows - I have to try/except import vlc, as my work Windows computer has 32bit python and 64bit vlc. I have put in a request for 64bit python. At home I only use Ubuntu and Linux Mint. The relative imports did not work on either machine, so I reverted to absolute imports.

PR #57, opened 2020-01-01
Develop

ccbogel:
add functionality to rewind 3 seconds, add time stamps and add and insert speaker names


ccbogel:
add functionality to rewind 3 seconds, add time stamps and add and insert speaker names

PR #58, opened 2020-01-01
Develop

ccbogel:


ccbogel:
improve manual transcribing and help files

PR #69, opened 2020-03-17
Develop

ccbogel:


ccbogel:
function to select av segment and assign it to selected text in text edit

PR #267, opened 2020-09-21
first attempt in expliciating how to run qualcoder on fedora

ilippert:
please let me know any change requests to the patch. Of course, if this patch messes up the readme file too much, I understand. In the latter case, maybe a different place for the specific linux-related install information may be helpful.


ccbogel:
@ilippert I wonder if the install.sh script could be modified to add the extra code needed to install ebooklib on Fedora ?

PR #269, opened 2020-10-08
  Update Citation APA style version number in README

menelic:
use correct version number in QualCoder citation


ccbogel:
Thanks @menelic 
Also, any further update from you on the critical bug?

PR #289, opened 2020-11-10
Update it.po

WPFilmmaker:
The translation of it.po should be completed. Two strings were left in English because I had no idea how to translate them. Typos were fixed.


ccbogel:
Thank you

PR #297, opened 2020-11-14
Update app_it.ts

WPFilmmaker:
v1 something like 80% done, some strings untrasnlated because I didn't know how. At 106 there is a typo in the word attribute, but I didn't know whether it was a typo or intended, let me know :)


ccbogel:
thank you

PR #298, opened 2020-11-15
Update app_it.ts

WPFilmmaker:
Translation should be done (apart from few lines I don't know how to translate without context). Check if this time translations are positioned in the right way :D


ccbogel:
Thank you, it looks good.

PR #353, opened 2020-12-19
Develop

ccbogel:


ccbogel:
Color selector marks used colors with asterix. Colors have tooltips with names of codes assigned to the color.

PR #377, opened 2021-01-30
macOS installation guide improvement proposals

tchakabam:


ccbogel:
Thanks very much. I do not use Mac so I rely on others for installation instructions.

ccbogel:
Thsnk you

tchakabam:
You're welcome :) 👍 Great job on providing this tool.

README.md
```diff
@@ -96,34 +96,62 @@ Install [Python3](https://www.python.org/downloads/) and [VLC](https://www.video
 
 Download Qualcoder-master Zip file and copy it into /Applications
```
tchakabam, line 97: @ccbogel you might want to fix the above instruction as it falls out of line with what the step-by-step instructions guide to (downloading and extracting the release zip being the last point).

Also, you might want to get rid of the duplicate wording above regarding installing Py & VLC? ;)

tchakabam:
@ccbogel I was doing this change actually as a friend of mine needed to use this tool, and not being an IT-wiz she got a bit lost with the instructions - like many people might be who are using this tool, since it is not for developers per se ;) 

So, I guess you can definitely help that with some more layout and fix wording in the docs, because there was partly also the issue, not the technical process on macOS to follow per se I would say :)

ccbogel:
Ich danke Ihnen, fuer ihre Vorschlaege.
Please see the updated instructions. 
Ich hoffe sie sind besser.

PR #380, opened 2021-02-03
add setuptool for native macOS app

kampfq:
works for me, further testing needed :)

just run `python3 setup.py py2app` on macOS

i also added multiplatform instructions, but never tested on windows


ccbogel:
Thanks very much :)  
I will further test soon.

PR #495, opened 2021-06-23
Improve setup.py and switch to relative imports

maralorn:
## Commit message

With the old setup.py I was not able to install the program so I tried
to fix the setup.py.

Now it creates a correct wheel, and creates a "qualcoder" executable
that will be installed into the /bin/ folder that setup-tools uses to
install software.

Using this install method the unqualified import didn‘t work, so I made
them relative.

I can not test this on windows, but I assume there has to be a way to
use setup.py to create an executable similar to the linux "qualcoder"
executable.

## Comment

I am not very certain about python best practices. Nevertheless I think this should be an improvement. I thought I should create this PR in the hope, that it might benefit you.

### DIFF
--- a/qualcoder/GUI/add_attribute.py
+++ b/qualcoder/GUI/add_attribute.py
@@ -32,7 +32,7 @@ import sys
 import logging
 import traceback
 
-from GUI.ui_attribute import Ui_DialogAddAttribute
+from .ui_attribute import Ui_DialogAddAttribute
 
 path = os.path.abspath(os.path.dirname(__file__))
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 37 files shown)


ccbogel:
Hello maralorn. Thank you very much for this. I will have to test, your fork on Windows first. 
Last time someone tried to change to relative imports QualCoder would not work on Windows. (by double click the qualcoder.py file). This is why I have tried to only use absolute imports. However, I see you have not changed the absolute imports within the qualcoder code. I will test and let you know.

mweinelt:
Should the relative imports indeed not work I'd recommend going with absolute, fully-qualified paths, that reference the package, like `from qualcoder.helpers import Message`.

Issue #500, opened 2021-06-29
Relative imports


ccbogel:
@maralorn  IMPORTANT
I cannot run the code from command line or via PyCharm using relative imports.

This is one reason why I  removed all the relative imports from a previous developer. I will do so again unless there is a good solution. I can use if else blocks around the imports (e.g. try importing relative then try importing absolute), but I do not think this is a great solution.

maralorn:
How are you trying to run the code?

maralorn:
I am sure that PyCharm has to have a way to work together with setuptools.

For running from command line, this is the way to go (only sure about this on linux systems):

Run `python setup.py install`. After that you can run `qualcoder` when the directory `setup.py` installs to is in your PATH variable. Otherwise run `path/to/qualcoder`.

If you want to immediately try out every change you can instead use `python setup.py develop`. (https://setuptools.readthedocs.io/en/latest/userguide/development_mode.html).

That way the `qualcoder` program is not a copy of your code but just a link into your working directory, so everytime you reexecute the program it uses the current state of your code.

ccbogel:
OK. I have python knowledge in some areas, but not in others.
I used this successfully to install via develop:, (and then qualcoder from the command line on Ubuntu 20.04):
sudo python3 setup.py develop

ccbogel:
Also. Does this mean the instillation instructions will have to change in the Readme? as this setup.py is now working better (thank you).
So for windows:
py setup.py install
Linux:
python3 setup.py install
Mac:
python setup.py install

ccbogel:
In PyCharm, I have marked qualcoder folder as the source root
![Screenshot from 2021-06-29 20-08-50](https://user-images.githubusercontent.com/5834638/123779896-fa312f00-d915-11eb-9ab6-d3ad202b45a7.png)

mweinelt:
Right, try to bulk replace `from .` with `from qualcoder.` to make these imports absolute. It is likely the better choice anyway.

ccbogel:
Danke, Ich wuerde probieren

maralorn:
My problem is, that I am also not a python expert and I feel a bit bad about sending you down a rabbit hole in which you could get lost.

But let’s try to tackle this one by one:




> I used this successfully to install via develop:, (and then qualcoder from the command line on Ubuntu 20.04):
> sudo python3 setup.py develop

Great!



> I find relative imports unclear. Do absolute imports not work with setup.py ?

[...]

PR #501, opened 2021-06-29
Fix import issues discussed in issue #500

fmagin:
qualcoder.py was renamed to `__main__.py` because Python will throw
incredibly confusing errors if a you have a `qualcoder.py` _file_ with the same name
as a _package_ (`qualcoder`), because it will try to import the file instead of the
package, if the the former comes first in the PYTHON_PATH.

This also allows running the module directly, i.e. `python -m
qualcoder` will now use python to resolve the the path of the module
qualcoder, and then basically run `python path/to/qualcoder/__main__.py`

The imports in this `__main__.py` are also changed to absolute imports,
because if the file is run directly, it isn't imported as a module, and
there is no module the imports could be relative to. The imports in the
other files can be left relative, because those files are imported
themselves, so a module to be relative to exists at that time.

### DIFF
--- a/qualcoder/__main__.py
+++ b/qualcoder/__main__.py
@@ -45,32 +45,32 @@ from copy import copy
 
 from PyQt5 import QtCore, QtGui, QtWidgets
 
-from .attributes import DialogManageAttributes
-from .cases import DialogCases
-from .codebook import Codebook
-from .code_text import DialogCodeText
-from .GUI.ui_main import Ui_MainWindow
-from .helpers import Message
-from .import_survey import DialogImportSurvey
-from .information import DialogInformation
-from .journals import DialogJournals
-from .manage_files import DialogManageFiles
-from .manage_links import DialogManageLinks
-from .memo import DialogMemo
-from .refi import RefiExport, RefiImport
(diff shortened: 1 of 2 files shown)


fmagin:
Actually, the `python -m qualcoder` doesn't work yet for some reason if I install this with `pip install ./` instead of `pip install -e` so this isn't fully fixed yet, will quickly check what's going on there.

fmagin:
@ccbogel I tested this on Linux, both inside PyCharm (when opening the git root directly, not just the qualcoder subfolder), and on the command line with `python -m qualcoder` both when installed in development mode with `pip install -e ./` and as a regular install with `pip install ./`

I can't test this on Windows, but the Python importing logic should be the same, so it _should_ also work. If you have any further questions about what caused this issue, and why @maralorn was nonetheless right to change the imports you had before please don't hesitate to ask before merging this, Python importing logic is a confusing mess.

ccbogel:
@fmagin Thank you very much.  Your explanation does help my understanding of the issue. I will merge now (2.40am my time) and test in the morning. I should have enough power in my Windows laptop to test on that OS.

PR #569, opened 2021-10-31
Dev

ccbogel:


ccbogel:
Speech to text functionality added

candideu:
> Speech to text functionality added

Hello! How can we get the speech-to-text working?

ccbogel:
@candideu Import the a/v file into qualcoder. QualCoder will make an empty text file associated with that a/v file. 
Then, while in Manage Files, double-click the a/v file to open it to view.
underneath the transcription re is a cogs button. Click this to open the speech to text option. I have made it work only with a blank transcription file.
pic1
.

candideu:
Thanks for the info!
I followed the steps, but received this error: 
image

ccbogel:
Hmm. that could be too tricky for me to fix. 
As I don't manage the pydub program. 
Do you have the program ffmpeg installed?
Then try re-running the above.

If the above does not work. Then separately use a speech to text service such as through google or google docs. Ad then copy and paste the text into the text transcript file via Manage Files then open the text file.

candideu:
I installed ffmpeg, and it works. Thanks for your help!
If I get the chance, I'll update the documentation.

I'm not sure if this would be helpful for this project, but Mozilla Deep Speech might also be a possible engine (used by the AutoSub project: https://github.com/abhirooptalasila/AutoSub)

ccbogel:
ok thanks for that information and updating the manual

Issue #589, opened 2021-12-16
missing package on fedora to compile


ilippert:
**Describe the bug**
fresh install of fedora 35; qualcoder does not compile.

**To Reproduce**
fresh install of fedora 35; try to compile.

**Expected behavior**
compilation should proceed.

**Screenshots**
this is the error message
```
Running PyQt5_sip-12.9.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-qu82msd5/PyQt5_sip-12.9.0/egg-dist-tmp-qio6cdzg
apiversions.c:20:10: fatal error: Python.h: No such file or directory
   20 | #include <Python.h>
      |          ^~~~~~~~~~
compilation terminated.
```

ccbogel:
Thanks ingmar

PR #590, opened 2021-12-16
Update install_fedora.sh

ilippert:
adding a) missing dependency, b) an automatic yes to the install command

to fix
https://github.com/ccbogel/QualCoder/issues/589


ccbogel:
Thanks Ingmar

PR #611, opened 2022-01-31
Update qualcoder.desktop

ilippert:
adding keywords that I often use to look for the programme; for specifications see e.g. https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html


ccbogel:
Thsnk you

Issue #625, opened 2022-03-20
While importing documents, clean up bad characters


ricardoamaro:
**Is your feature request related to a problem? Please describe.**
An imported text sometimes generates special characters. Those characters are not escaped or cleaned up during import and break the autocoding with line endings.

**Describe the solution you'd like**

Import a PDF book.

These characters appear:

![image](https://user-images.githubusercontent.com/384962/159162521-ae52c842-b81a-424d-8ab2-253e76c99aee.png)

```
    e picture they paint is compelling. ey describe how eﬀective IT
```

In my opinion, we could clean up the text on import.

**Describe alternatives you've considered**
Text should be clear from special characters on import to insure clear reading and that things do not break.

**Additional context**
Already mentioned in https://github.com/ccbogel/QualCoder/issues/623

ccbogel:
I'm not sure how to best do this, apart from what has been started by adding those unicode replace characters in mange_files.load_file_data method.

ccbogel:
I will close this for now as  I have no solution to the problem. but raise it with me when bad characters arise. So  I can do a fix as needed.

PR #627, opened 2022-03-20
Fix Pdfminer recognizing invalid unicode characters.

ricardoamaro:
Addresses While importing documents, clean up bad characters https://github.com/ccbogel/QualCoder/issues/625


ccbogel:
great thank you!

PR #647, opened 2022-05-09
Dev

ccbogel:
heatmaps for files and cases added to Charts


ccbogel:
This dev update was only for the Reports > Charts options .
Now can do heatmap charts.
I had to update 3 different files to make this happen, so I had to make a dev branch before merging all of these.
with regards Colin
________________________________

PR #746, opened 2023-01-20
add native os styling as option

kampfq:
native styling means that the ui is dark/light depending on the system preferences
the ui looks also more native on windows and macOS
when native is active use system default menubar on macOS 
removed the dynamic low res appicon on macos because it comes with the build binary


ccbogel:
This looks OK. But dark/light might be occasional visual issues: zum beispeil: waveform color; text boxes displayed in the graph.

PR #791, opened 2023-04-24
Correct typo in report_codes.py

jbkieffer:
On line 1345, one of the `ifnull` clauses was spelled `ifnul` and caused an error in the Coding Reports after the most recent updates. Correcting this typo in my local copy solved the error, so offering it back as a PR.


ccbogel:
Thank you very much. Ironically, I added the ifnull everywhere to reduce errors.

PR #804, opened 2023-07-18
BUGFIX: spurious + sign in the calculation of the documents import path.

### LINKED ISSUE #803: REFI-QDA import error: not importing documents
**Describe the bug**

The import of a REFI-QDA project fails to include the documents.

**To Reproduce**

- Select a .qdpx project file with documents 
- Once the import finishes, go to manage documents. Nothing will be there.

**Expected behavior**

The documents included in the project should be there.

**Screenshots**

The console output is:

  File "/Users/user/Projects/QualCoder/qualcoder/refi.py", line 1319, in load_pdf_source
    destination = os.path.join(self.app.project_path, +"documents", name)
                                                      ^^^^^^^^^^^^

**Desktop (please complete the following information):**
 - OS: macOS 12.6.7
 - Python 3.11

melegati:
Fixes #803.

An extra + was present in a call to os.path what led to an error that prevented the proper import of REFI-QDA file.

### DIFF
--- a/qualcoder/refi.py
+++ b/qualcoder/refi.py
@@ -1316,7 +1316,7 @@ class RefiImport:
         media_path = "/docs/" + name  # Default
         if path_type == "internal":
             # Copy file into .qda documents folder and rename into original name
-            destination = os.path.join(self.app.project_path, +"documents", name)
+            destination = os.path.join(self.app.project_path, "documents", name)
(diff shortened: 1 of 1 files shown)


ccbogel:
@melegati Thank You! I recently began using the join method, and missed that plus sign.

PR #831, opened 2023-10-24
fix: environment variable setup for windows this will resolve all the dependencies issues

Ayushpanditmoto:
… dependencies issues


ccbogel:
thank you

ccbogel:
@Ayushpanditmoto HI. That updated source command did not work. SO i changed back to this:
py -m venv env
env\Scripts\activate

PR #841, opened 2023-10-28
Update README.md

shraddha761:
I update readme with badges which make it look more attractive


shraddha761:
@ccbogel please review

PR #869, opened 2023-12-13
Dev

ccbogel:
Removed dependency on external python wordcloud module.
The dependency did not work well on Fedora, and could not be packaged with pyinstaller.

Replaced with a simpler wordcloud class. (Does not use shape masks. Does no apply text rotations. Minimises but does not completely remove text overlaps).


ccbogel:
Removed dependency on external wordcloud

Issue #849, opened 2023-10-29
Requirements.txt instead of listing all modules


Paulie-Aditya:
The readme states that I need to download so and so modules manually. This looks shabby and hard to understand for users too.

**Describe the solution you'd like**  
A Requirements.txt file so that people can easily install the required packages with a simple line and get on with developing using this repo.

ilippert:
For which operation OS do you like this?

ccbogel:
added requirements file

PR #873, opened 2023-12-15
seperate out the required packages

ilippert:
towards solving https://github.com/ccbogel/QualCoder/issues/849 I separated out the (to me known) required packages


ilippert:
Sorry, these are a range of commits, doing different things; I did not intuitively manage to create different merge requests for these different commits. B ut will try to do that if you prefer that.

PR #892, opened 2024-02-28
Patch 14 includes a cleaning of the install script, moving python-vlc to rpm fusion repo install

ilippert:
still not entirely working, but cleaned up.


ccbogel:
did this fix the problem?

PR #899, opened 2024-04-11
Improved collaboration functions using a shared network drive

kaixxx:
See discussion here: https://github.com/ccbogel/QualCoder/discussions/894

### Changes:

**Adding a lock file for concurrent access protection (main.py)** 
- The lock file is created on opening a project and deleted when it is closed. 
- While a project is being used, a regular heartbeat (timestamp) is written to this file every 5 seconds. If QualCoder encounters a problem writing this heartbeat (e.g., because the network is down), the user is notified and advised to close the project.
- If QualCoder tries to open a project and finds a lock file in it, it checks the timestamp: If the timestamp is older than 30 seconds, QualCoder assumes that the original owner process died or has lost the connection to the file. In this case, QualCoder ignores the lock and opens the project like normal (with a notification). If the timestamp is younger than 30 seconds, QualCoder notifies the user that the project is locked and cannot be opened.

To test, start two instances of QualCoder (even on the same computer) and try to open the same project.

**"atomic commits"**
This ensures that during complex operations resulting in a lot of changes to the database, either all changes are applied at once ("atomic") or - in case of an error - the database is rolled back to its previous unchanged state. This should significantly reduce the risk of the database becoming inconsistent as a result of a network error during such an operation. 
Procedure: Enclosing the operation in a try .. except block with only one call to commit() as the very last step. The except-clause contains a rollback to the state before the operation. 
Applied to the following functions:
- `merge_codes()` and `merge_categories()` in code_text, code_pdf, view_av, and view_image
- `edit_textfile.execute()`
- `add_attribute()` in manage_files.py

Feel free to ask if you have any questions!


kaixxx:
A quick note: If you accept these changes, you may want to make a new release. If so, please give me a little bit of time to update my AI-enhanced version too. My users will be notified about your new version (I've left the code in there). So I would like to have an updated version ready at that time also. 

I'm still not sure when the AI-functionality will be ready to be incorporated in the main version of QualCoder - might be next week (unlikely), might be in a couple of weeks. I have a few people working with these functions in their research and giving me feedback. There is still some work to do.

kaixxx:
I have written a new page for the wiki explaining the different approaches to work in a team with QualCoder. I would suggest replacing the note on the first page of the wiki with a link to this new page. 
Strangely, GitHub is not allowing pull requests for wikis. So I'll post my suggested text here:

+++++

# Working Together in a Team

There are several strategies for organizing collaborative work on the same project in QualCoder. Choose whichever best fits your methodological approach and specific working environment.

## Sequential Editing of the Same Project
This approach can be applied to both inductive and deductive coding. It is based on the principle that all team members work on the same project files, albeit not simultaneously.
### Working With a Network Drive
- The simplest method for sequential editing involves saving your project on a network drive accessible to all team members. **Caution: Cloud-based shared folders such as OneDrive, Dropbox, etc., are unsuitable for this purpose.** You must use a genuine network drive where QualCoder can directly write, such as on your university's file server. Moreover, a very stable network connection is crucial to avoid errors.
- QualCoder can now automatically manage concurrent access. Attempting to open a project that someone else is already using results in an error message. This message also displays the username of the person currently working on the project.
### Coordinating the Sequence of Work Manually
- You can do this via email so that each person, upon completing their tasks, mails the zipped project folder to the subsequent individual on the list.
- You can also work sequentially on the same computer, using the same project folder.
- If your coordinate the work manually, using a cloud-sharing service like OneDrive or Dropbox to transfer the files is also an option. However, it's critical that a) under no circumstances should two people access the same project simultaneously and b) the project must be fully synced with the cloud service *before* passing the work to the next individual. The succeeding person must also ensure they have downloaded the latest version from the cloud before beginning their tasks. Be careful: Project versions getting out of sync can lead to complex issues that are challenging to rectify.
- Please do not use REFI-QDA Project export as a routine approach for collaboration when all coders are using QualCoder. Use one of the other options described here instead. If you need to open a project in different software (e.g., Taguette, NVivo, Atlas.ti, MAXQDA, Dedoose, Quirkos), only then should you use the REFI-QDA Project export. It is important to recognize that project features can be lost when transferring a project between qualitative software using the REFI-QDA Project export.
- Handle coder-names carefully (see below).

## Using a Master Project: The 'Coding Separately Then Merging' Approach

This strategy allows your team to work simultaneously to some extent. It is most effective if you use a deductive coding approach, where the code system doesn't change much during the analysis. If you use it with inductive coding, be prepared to spend some time cleaning up and consolidating the code system when merging the work of different team members.

Follow these steps:
- Create a master project with the code system already defined (if applicable).
- Ideally, import all empirical documents at this stage. Avoid linking to external files as this complicates the distribution of the project.
- Zip your project folder, including all subfolders and files, and send it to your collaborators (or use an USB-stick).
- Decide on a strategy for handling coder names in your team, as explained in the notes below.
- Now, each team member can unzip their local copy of the project and work on it independently, adding codes, codings, memos, etc. **However, do not edit the text of the empirical documents themselves, since this will cause problems during the merging process later.**
- Finally, gather all the local copies and sequentially merge them into your master project, using the process described here: https://github.com/ccbogel/QualCoder/wiki/12-Imports-and-exports#merge-qualcoder-projects

## Important Considerations Regarding Coder Names
No matter which strategy you use to organize the collaborative coding process, you must decide if you want to use individual coder names or not. 
- In most cases, you want to have a unique name for each coder on your team so that individual contributions can be identified even after merging the project together. To achieve this, please check the current coder's name in Files>Settings before doing any work.
- The coder name is 'default' when first using QualCoder. When opening another existing project, please make sure the coder name is set to the name you want to use by checking and updating the Settings before you work on your project. You can either select an existing coder name from the dropdown list or enter a new one and click the Apply button.
- Be aware that in the coding workspaces (text, images, A/V), only the coded segments of the currently selected coder in Settings will be visible. So, if nothing shows, change the coder name using the dropdown box in Settings.
- This also indicates a potential downside of using individual coder names: You cannot easily see what others have already coded in your document (applies only to the sequential work process described above). If you need that, use the same name for all coders throughout.

ccbogel:
@kaixxx OK thank you for your work on this. If the AI is no ready yet,  I will wait one or more weeks first. I will create a draft page for the Wiki.
It might be that the wiki was a problem a long time ago. I had a malicious person replace links with links to malicious sites. So I then restricted editing. I will see if I can change those settings.

kaixxx:
In code_text.py, applied atomic commits to:
auto_code()
code_sentences()
button_autocode_surround()
undo_autocoding()

One problem in code_sentences(): There was already a try...except block (3743-3757) around the main database operation that catches every error but keeps the function running. This counteracts the 'atomic commit' try...except block that I've build around it because no error will reach it. Reraising the error in the inner try...except block would solve this. What do you think? 


In view_graph.py, applied atomic commits to
- save_graph()
- delete_saved_graph()


Regarding merge_project: This function is currently protected by an automatic backup. Even though this takes a little bit of time, it's the better approach in this particular case IMHO. Merging projects can include a lot of file operations. A clean rollback would need to address not only the database, but also the files, which is complicated to implement.

kaixxx:
I have access to the wiki, thank you!

ccbogel:
I have made those wiki edits.

PR #909, opened 2024-05-15
Fix settings dialog: Selecting current style in languages other than …

kaixxx:
…English

In languages other than English, the current style was always reset back to 'original' on opening the settings dialog because QualCoder expected the English name, not the translated one.


ccbogel:
thank you

PR #925, opened 2024-07-17
Fixing a few more file encoding issues

kaixxx:
- previous_project_paths can also contain non-ascii characters
- The lock file stores the name of the current user, so it seems a good idea to make this UTF-8 also.


ccbogel:
Thank you

Issue #928, opened 2024-08-06
Ideas for Improved Usability


kaixxx:
While working on my AI-enhanced version of QualCoder (which is close to completion) I discovered some potential enhancements for the usability of QualCoder that I would like to discuss. I can make pull requests if you like my suggestions.

1) **Main Window Geometry Save & Restore:** Currently, only the width and height of the main window are saved in config.ini, omitting the position and the crucial maximized/full-screen state. *I recommend storing and restoring the entire geometry of the app.* The typical method in PyQt involves using the `saveGeometry()` function of the main window. This would replace `mainwindow_w` and `mainwindow_h` in config.ini.

2) **Native Light/Dark Mode:** Since version 3.5, PyQt6 automatically switches between light and dark mode based on the system's settings (only affects the 'native' style in QualCoder). This feature works on Windows, macOS, and the Gnome desktop. For more information, visit: https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5. 
The result looks quite nice, imho (see below). *I would suggest making the 'native' style the default for new users of QualCoder.* This way, people will feel more at home when they first start the app, as the styling seamlessly integrates with their system settings. If they want to try out some of the fancier styles, they can still do so by switching in the settings.

![QualCoder_dark](https://github.com/user-attachments/assets/aab0ab94-701d-4f0d-ba9e-0fcaef4f9fc9)
Light mode (needs a little refinement at the toolbar):
![QualCoder_light](https://github.com/user-attachments/assets/494240aa-2d81-4fac-bb3e-3ac5d25085b1)

3) **Fusion Style on Windows**: Qt recommends using the built-in "Fusion" style as a base for any further styling on Windows, since it has the best support for native dark mode (see the link above). I have found this to be true as well and would *suggest making "Fusion" the default on Windows.* (The screenshots above are already using the Fusion style.)

4) **More Structured Error Message Dialog:** I've found it extremely helpful to include trace logs in the error messages in QualCoder. (I’m probably going to steal this idea for my other app...) However, in the case of the libraries I use for AI integration, these trace logs can become quite lengthy due to the many layers of abstraction. This results in the error message dialog expanding so much that it exceeds the screen size, making the confirmation button unreachable. *I would suggest creating a new error message dialog with the actual message at the top and the trace log below in a scrollable textbox of a fixed size.* 
[...]

PR #931, opened 2024-08-08
Ideas for Improved Usability II

kaixxx:
As discussed here: https://github.com/ccbogel/QualCoder/issues/928
Continuation of: https://github.com/ccbogel/QualCoder/pull/929

The new error dialog looks like this: 

The 'Copy' button copies the error message and traceback to the clipboard for easy reporting.

One importand thing to consider: `sys.excepthook` is a global hook for the current instance of the python interpretor. It only needs to be set once for the whole app. This is now triggered by the line `qt_exception_hook = UncaughtHook()` in __main__.py, right after setting up the logger. **ALL other assignments to sys.excepthook in the whole app must to be removed.** Otherwise, the new error handling routine would be overwritten by these old methods. For the basic concept see: https://timlehr.com/2018/01/python-exception-hooks-with-qt-message-box/index.html 
I can remove the old exception_handlers but wanted to get your approval before making such a widespread change to the current codebase. 

For the time beeing, I have only replaced the exception_handler in __main__.py. To test this, you can follow these steps:
- Start the app and open a project, but do not open any dialogs etc. (loading such a dialog would reinstate the old exception_handler)
- Click on Project > Create New Project, then CANCEL the file dialog.
- Now click on Project > Project Memo. You should see the new error dialog saying that a 'NoneType' object has no attribute 'cursor'. The reason is a small bug in new_project(): The new App-object is already created before the file dialog is opened. I can fix this later.

Final note: I've also added a small script called 'rebuild_ui.py' that I use to automatically recompile the UI-files after making changes in Qt Designer. You can ignore it if you want.


ccbogel:
@kaixxx  yes ok. One thing I am thinking is that the full traceback error should go to the log file.
Usually having the short immediate error is fine, but there are times when the full track is useful to see the preceding steps that initiated the error.

PR #935, opened 2024-09-13
Make columns movable and resizable in the code tree widget

seelebrn:
Could be made better, but I set a starting fixed size of 400 to avoid having the "Name" column too small by defaut.


ccbogel:
thank you

PR #936, opened 2024-09-13
Bulk Renaming of Files

seelebrn:
Added a context menu option to bulk rename files
Ctrl-Q is not a satisfying option, but Ctrl-A is already taken.

### DIFF
--- a/qualcoder/manage_files.py
+++ b/qualcoder/manage_files.py
@@ -172,7 +172,7 @@ class DialogManageFiles(QtWidgets.QDialog):
         self.ui.tableWidget.cellDoubleClicked.connect(self.cell_double_clicked)
         self.ui.tableWidget.setContextMenuPolicy(QtCore.Qt.ContextMenuPolicy.CustomContextMenu)
         self.ui.tableWidget.customContextMenuRequested.connect(self.table_menu)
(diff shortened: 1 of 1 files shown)


ccbogel:
HI. thanks for the pull request. I won't be using it as I had already begun working on this, based on your code, and will update very soon today.
I had added extra error catching around the name update in the database.
I cannot enact the MultiSelection table selection mode. I did do this years ago, and while I cannot remember the details, I do remember this it did cause some problems/errors at the time. Instead, I have made it so that all visible rows are updated. Hidden rows are not updated. So you could select based on an attribute or name value
I have also added a 'hide rows like' option in the context menu to hide rows containing a value

ccbogel:
HI. thanks for the pull request. I won't be using it as I had already begun working on this, based on your code, and will update very soon today.
I had added extra error catching around the name update in the database.
I cannot enact the MultiSelection table selection mode. I did do this years ago, and while I cannot remember the details, I do remember this it did cause some problems/errors at the time. Instead, I have made it so that all visible rows are updated. Hidden rows are not updated. So you could select based on an attribute or name value
I have also added a 'hide rows like' option in the context menu to hide rows containing a value

PR #937, opened 2024-09-19
fix font settings

kaixxx:
The font stylesheet in "init_ui()" had a problem of not enclosing the font name correctly with quotation marks. But since we set the font already in "__init()__", we can delete it here.


ccbogel:
Thanks I was updating with fStrings.

kaixxx:
> I was updating with fStrings

I've seen that, much cleaner

PR #948, opened 2024-10-17
Better docstrings in code_text.py

kaixxx:
Described the AI related functions a little better


kaixxx:
Also added some minimal version requirements for langchain.

ccbogel:
Thanks for these updates and the docstrings

PR #956, opened 2024-10-29
UI improvements to AI Search and Chat

kaixxx:
I've added some improvements (see commits) based on our email discussion - most notably, marking the 3 important decisions in the AI Search window more clearly: 

grafik

I hope this makes the dialog more intuitive to use (the same dialog is used in AI Search and the various chats).
The new screenshot is also in the wiki now.


ccbogel:
yes that looks good thank you

PR #968, opened 2024-11-12
Small updates in AI features

kaixxx:
I also fixed a small error in QualCoder:
A text doc created in QualCoder has a "mediapath" of None. This was causing an error trying to access this field in the process of deleting the doc from the database.


ccbogel:
thank you

PR #979, opened 2024-12-14
Onefile spec file for testing

kaixxx:
I've added an additional spec file for compilation in onefile mode. However, I would not recommend using it. Starting this onefile app will take around 20 seconds, compared to 6 seconds with a onedir app. The normal qualcoder.spec creates such a onedir app.

### DIFF
--- a/qualcoder_onefile.spec
+++ b/qualcoder_onefile.spec
@@ -41,21 +41,17 @@ a = Analysis(
     cipher=block_cipher,
     noarchive=False,
 )
-pyz = PYZ(a.pure, a.zipped_data, cipher=block_cipher)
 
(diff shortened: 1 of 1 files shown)


ccbogel:
Yes, sure. Although the one file mode would be easy for users to download from the releases page. but all options can be provided. along with the wait time warning. I thin Nvivo takes quite a few seconds to start up also.

kaixxx:
> the one file mode would be easy for users to download
Yeah, the onedir version must be combined with an installer. We can make one for Windows and I hope that Fredrik from Uppsala will make one for macOS. On Linux, however, the onfile version might be the easiest solution.

PR #981, opened 2024-12-16
Adding Line Numbers

kaixxx:
I know I should stop adding features and concentrate on the release ;)
But there was one thing several people complained about in the workshops that I recently gave: missing line numbers.
In Europe, it is quite common to reference passages in empirical documents with their exact line numbers. Commercial packages like MAXQDA offer this feature, QualCoder did not. So, I've added something to code_text: 

The implementation is loosely based on https://nachtimwald.com/2009/08/15/qtextedit-with-line-numbers/

I've also added line numbers to references in AI chats.

What do you think? I've tried to make the numbers as unobtrusive as possible.


ccbogel:
ok, will pull the request and see :)

ccbogel:
Yes I like it. Great work. p.s. I have been working on extending the code image functions - again not really needed, but might be handy.
One thing - and I can't test this right now, what happens if the document is very large, and when right clicking on opening the file, the user chooses to open the text in 50,000 character blocks. so 50000 char chunks of text are shown, will this affect the line numbering?
Also, what else did the workshop people complain about? - gives ideas for more functions

kaixxx:
> what happens if the document is very large, and when right clicking on opening the file, the user chooses to open the text in 50,000 character blocks

I forgot about that option, now rectified: https://github.com/ccbogel/QualCoder/pull/982

> what else did the workshop people complain about? 

Not much. But I'll make a list of small things that could be improved to make the user interface more intuitive. I've also asked my students to take some notes of what could be improved. But nothing seems very urgent, as far as I can tell.

ccbogel:
ok, thank you

PR #987, opened 2025-01-08
Fixing incompatibilities in AI memory between Windows and Mac

kaixxx:
I have bought a used Mac mini (M1 from 2020) for testing. QualCoder works fine, but there was one issue: 
If you open a project created on Windows on a Mac, the faiss index representing the AI memory is not compatible, which will through an error. The reason is that on Windows, the index was optimized for hardware acceleration based on AVX2, which is not available on newer Macs with Apple Silicon. 
I had to downgrade the `faiss-cpu` package to version 1.7.2 because newer versions would not allow to turn AVX2 off (a known issue: https://github.com/langchain-ai/chat-langchain/issues/61#issuecomment-1914426775). 
I also added error handling in case the faiss index needs to rebuilding.

### DIFF
--- a/qualcoder/ai_chat.py
+++ b/qualcoder/ai_chat.py
@@ -627,7 +627,10 @@ data collected. This information will accompany every prompt sent to the AI, res
                 self.ui.progressBar_ai.setRange(0, 0)  # Starts the animation
         # update ai status in the statusBar of the main window
         if self.app.ai is not None:
-            self.main_window.statusBar().showMessage(_('AI: ') + _(self.app.ai.get_status()))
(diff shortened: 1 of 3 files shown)


ccbogel:
OK.
Alternatively, perhaps if no other  soulution - to warn end users of using only one OS .. ? maybe ..?

kaixxx:
I have found a solution: We need to set a particular environment variable to stop faiss from using AVX2, even with the current version of the lib. It was hard to find, though, this option is not very well documented. 
I may have to adjust the error handling a little. After that, I make this PR final.

PR #992, opened 2025-01-14
Some improvements before release

kaixxx:
I have fixed some minor issues and also worked on small things that were still on my list:
- Added context menu to AI chat list like you requested
- Added copy & paste to AI prompt library, so people can share prompts more easy
- Added icons to main tabs to distinguish them better from the equally named menus above (a suggestion from one of my students):
grafik

It's purely visual. I can remove the icons if you don't like them.


ccbogel:
thank you

PR #995, opened 2025-01-21
Updated German translation (.po)

kaixxx:
The automatic translation was not bad, but there were also some funny mishaps, e.g. "AI Chat" translated into "AI Geschwätz" - meaning "chatter" or "blabber"... 
I'll also try to look at the .ts file from Qt, so that everything is consistent.


ccbogel:
OK thank you. Haha - blabber.
Unfortunately I am unable to update any ts files. I used to be able to do these on my older Linux Ubuntu OS., but now nothing seems to work on my Windows machine.

Issue #1024, opened 2025-03-07
Get LLM model info through API


VRehnberg:
**Is your feature request related to a problem? Please describe.**  
I'm running the AI part by serving local models with `vllm serve` and then I've written that entry in the config like e.g.:

```
[ai_model_vLLM_SmolLM2-1.7B]
desc = A small model you host locally with vLLM.
access_info_url = https://huggingface.co/HuggingFaceTB/SmolLM2-1.7B-Instruct
large_model = HuggingFaceTB/SmolLM2-1.7B-Instruct
large_model_context_window = 2048
fast_model = HuggingFaceTB/SmolLM2-1.7B-Instruct
fast_model_context_window = 2048
api_base = http://localhost:8000/v1
api_key = EMPTY
```

**Describe the solution you'd like**
Instead of specifying all this in the context it would instead look something like:
```
[ai_model_localhost]
desc = A self hosted model available through localhost.
access_info_url = https://huggingface.co/
api_base = http://localhost:8000/v1
api_key = EMPTY
```

And then in the AI settings menu you would instead only select provider "OpenAI", "blabladoor", "localhost". And when that is selected you can populate a list of available models and at the same time get the context window.

You should be able to get this information through the OpenAI API standard I believe. With vLLM you can at least query for the following information:

```bash
$ curl http://localhost:8000/v1/models
{"object":"list","data":[{"id":"HuggingFaceTB/SmolLM2-1.7B-Instruct","object":"model","created":1741351972,"owned_by":"vllm","root":"HuggingFaceTB/SmolLM2-1.7B-Instruct","parent":null,"max_model_len":2048,"permission":[{"id":"modelperm-1293a74b50394cb1aa26b15b5de4462b","object":"model_permission","created":1741351972,"allow_create_engine":false,"allow_sampling":true,"allow_logprobs":true,"allow_search_indices":false,"allow_view":true,"allow_fine_tuning":false,"organization":"*","group":null,"is_blocking":false}]}]}
```

**Describe alternatives you've considered**  
If you agree this would be convenient then I could see if I could write up a PR.

kaixxx:
That's an interesting idea. What I like about this in particular is that it would allow unexperienced users to switch to newer models which are published after the QualCoder release without messing with the config file. So if you want to develop something in that direction, I would encourage you.

A few remarks: 
[...]

PR #1036, opened 2025-04-03
Fine grained control over choosing an AI model in the GUI

kaixxx:
In response to https://github.com/ccbogel/QualCoder/issues/1024, I've now added a new section with "Advanced AI Options" in the settings dialog where one can choose among all the available models from the provider (OpenAI, Blablador, Google...):

I made it so that the advanced options are only visible after clicking the button, since this is for users who know what they are doing.


ccbogel:
exciting

PR #1042, opened 2025-04-10
Allowing to select text with the keyboard in code text

### LINKED ISSUE #1040: Keyboard arrows moving cursor in text
In text editors it is common to be able to use keyboard arrows to move around - otherwise coding becomes very tedious with the mouse. I am currently only able to scoll with the up and down keys, but not move the cursor. Well, to be fair, there is no cursor. Is this possible to implement? 
When I do try Alt + LeftArrow I get the error

```python
  File "C:\Users\py128\NIFU\Metode - General\Programvare\QualCoder\QualCoder-master\qualcoder\code_text.py", line 2234, in eventFilter
    self.shrink_to_left(code_)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^

  File "C:\Users\py128\NIFU\Metode - General\Programvare\QualCoder\QualCoder-master\qualcoder\code_text.py", line 2292, in shrink_to_left
    if code_['pos1'] <= code_['pos0'] + 1:
       ~~~~~^^^^^^^^

TypeError: 'NoneType' object is not subscriptable
```

Using Windows 11 and QualCoder 3.7

kaixxx:
`self.ui.textEdit.setTextInteractionFlags()` gives a more fine-grained control over what users can do with a textEdit. This allows to make the textEdit read only while still maintaining a visible cursor that can be controlled with the keyboard (to scroll, select text, etc.). 

I had to adjust the eventFilter a little so that within a coded (marked) section, shift + left etc. keep there original function without also changed the selected text. 

Fixes #1040

### DIFF
--- a/qualcoder/code_text.py
+++ b/qualcoder/code_text.py
@@ -193,7 +193,10 @@ class DialogCodeText(QtWidgets.QWidget):
         self.ui.textEdit.setAutoFillBackground(True)
         self.ui.textEdit.setToolTip("")
         self.ui.textEdit.setMouseTracking(True)
-        self.ui.textEdit.setReadOnly(True)
+        self.ui.textEdit.setTextInteractionFlags(
+            Qt.TextInteractionFlag.TextSelectableByMouse | 
+            Qt.TextInteractionFlag.TextSelectableByKeyboard  
+        )        
         self.ui.textEdit.installEventFilter(self)
         self.eventFilterTT = ToolTipEventFilter()
         self.ui.textEdit.installEventFilter(self.eventFilterTT)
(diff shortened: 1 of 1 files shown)


ccbogel:
Oh very nice. 
Where did you develop your coding skills?
Mine has been trial and error over years, plus some formal education.

kaixxx:
> Where did you develop your coding skills?

TBH, I've found this more or less by accident. I looked around in the QDesigner for a possibility to enable the cursor in the text view, and found the flag "TextSelectableByKeyboard" - which was already set, but seemed to have no effect. So I dug a little deeper into the TextInteractionFlags and tried a few combinations. 

On a more general level, ChatGPT has actually been very helpful over the last months in teaching me new tricks in Python and especially PyQt. I use it primarily as a brainstorming tool, discussing a certain problem and possible ways to solve it. I rarely use the suggested code directly. But I get ideas about certain concepts or options in Qt that I can then research via Google, etc. Besides, using ChatGPTs ability to quickly write a minimal example app to test a certain approach has proven to be very helpful as well.

PR #1045, opened 2025-04-15
Update app_es.ts

Salomon-Mazatlan:
The Spanish equivalent of 'Prompt' has been updated to 'instrucción' (singular) or 'instrucciones' (plural), rather than using 'aviso' or the incorrect 'alviso'.

### DIFF
--- a/qualcoder/GUI/app_es.ts
+++ b/qualcoder/GUI/app_es.ts
@@ -578,12 +578,12 @@ es-AU</translation>
     <message>
         <location filename="ui_ai_edit_prompts.py" line="195"/>
         <source>AI Prompts Library</source>
-        <translation>Biblioteca AI Alviso</translation>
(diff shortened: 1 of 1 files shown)


kaixxx:
Thank you very much! Could you please also take a look at the translation of the message strings here: https://github.com/ccbogel/QualCoder/blob/master/qualcoder/es.po It would be nice to have them consistent.

Salomon-Mazatlan:
> Thank you very much! Could you please also take a look at the translation of the message strings here: https://github.com/ccbogel/QualCoder/blob/master/qualcoder/es.po It would be nice to have them consistent.

Thank you very much, I will be reviewing it. Likewise, I am still checking the consistency with the terms IA and AI (for their Spanish acronyms).  :)

ccbogel:
Thank you for this.

PR #1046, opened 2025-04-15
Update es.po

Salomon-Mazatlan:
The acronym "AI" was changed to "IA" (for its Spanish equivalent) for greater consistency in the Spanish version, just as "Prompt" and "Indicadores" were replaced with "Instrucción (singular)" and "Instrucciones (plural)."

Additionally, the following lines (with their respective translations only) were modified, as they changed the letter indicating directionality (Left and Right):

2698 - "L rotate clockwise\n"
2699 - "R rotate anti-clockwise\n"
6737 - "L rotate clockwise\n"
6738 - "R rotate anti-clockwise\n"
7582 - "Rotate clockwise R"
7586 - "Rotate counter-clockwise L"

"Sentido horario" was replaced with "Girar 90° grados a la derecha," and "Girar en sentido antihorario L" was replaced with "Girar 90° grados a la izquierda."

### DIFF
--- a/qualcoder/es.po
+++ b/qualcoder/es.po
@@ -43,7 +43,7 @@ msgstr ""
 "Las nuevas funciones mejoradas de IA en QualCoder necesitan una "
 "configuración adicional. ¿Quieres habilitar la IA e iniciar la "
 "configuración? También puede hacerlo más tarde iniciando el Asistente de "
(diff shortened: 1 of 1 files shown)


ccbogel:
Thank you. yes please change anything you see is incorrect. We know how much work is involved in trawling through the translation files, so it is much appreciated.

PR #1050, opened 2025-04-25
Update es.po

Salomon-Mazatlan:
Grammar syntax corrections and translation of the new options were made (new lines at the end).

I will be on the lookout for the new additions. A big hug from Mazatlán, Sinaloa, México!


ccbogel:
thank you

Issue #1054, opened 2025-05-07
Implementing a test-suite


VRehnberg:
**Is your feature request related to a problem? Please describe.**  
There are many great new features added to QualCoder, but bleeding edge is currently quite flaky. Even running actual releases throws the occasional error message.

**Describe the solution you'd like**  
I believe adding a test-suite with unitit testing or more and adding this to the github CI could probably find most of these errors.

To lower the threshold I suggest starting with just some very simple unit tests in the CI and then adding some more with time to increase the test coverage. I personally have good experience with the pytest package but the built in unittest package works great as well. If you go with pytest, there is a pytest-qt project that seems quite well maintained.

To get started [CodeRefinery has a guide](https://coderefinery.github.io/testing/) ((video)[https://www.youtube.com/watch?v=-Rwu5CMtY5I&list=PLpLblYHCzJACpOmIzO8TywjtfYD7_d93H&index=11]) or if you prefer I could probably create a basic PR that you could build on.

**Describe alternatives you've considered**
More descriptive commit messages would help as well.

ccbogel:
OK thank you,  I had avoided this as it seemed too hard to do.

ccbogel:
So now I am getting lots of emails like this:

![Image](https://github.com/user-attachments/assets/8cbe6e61-fc69-4e28-8302-a58714be2357)

I updated some details in the **test_qualcoder.py** file to make the database structure current, and copied the settings dictionary from another project.

VRehnberg:
If it's the same as Gitlab, you'll get an email per commit on master that fails and then an email for the commit that fixes it. And after that no emails as long as everything is fine.

The email preferences is probably configurable https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#configuring-your-watch-settings-for-an-individual-repository

Let's see if I have some time this afternoon to fix the current errors.

VRehnberg:
Submitted a PR that solved the two failing tests.

ccbogel:
OK thank you, Im learning from what you are doing.

PR #1057, opened 2025-05-12
Fix tests

VRehnberg:
Solves #1054

I'm actually only using one of the example config.ini files, but I figured I'd include both with and without AI for that specific version to make it easier for future testing.

### DIFF
--- a/tests/fixtures/config-ai-ec4c0559.ini
+++ b/tests/fixtures/config-ai-ec4c0559.ini
@@ -1,0 +1,119 @@
+[DEFAULT]
+backup_num = 5
+codername = default
+font = DejaVu Sans
+fontsize = 12
+docfontsize = 12
+treefontsize = 12
+directory = /cephyr/users/vikren/Alvis
+showids = False
+language = en
+backup_on_open = True
+backup_av_files = True
+timestampformat = [hh.mm.ss]
+speakernameformat = []
+mainwindow_geometry = 
+dialogcodetext_splitter0 = 1
+dialogcodetext_splitter1 = 1
+dialogcodetext_splitter_v0 = 1
+dialogcodetext_splitter_v1 = 1
+dialogcodeimage_splitter0 = 1
+dialogcodeimage_splitter1 = 1
+dialogcodeimage_splitter_h0 = 1
+dialogcodeimage_splitter_h1 = 1
+dialogreportcodes_splitter0 = 1
+dialogreportcodes_splitter1 = 1
+dialogreportcodes_splitter_v0 = 30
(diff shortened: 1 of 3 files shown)


ccbogel:
OK thank you - I have read through the changes you have made to learn from it.

kaixxx:
Thank you for implementing these test, Viktor! 
One thing worries me a bit: The handling of the config.ini is already quite complicated, and making changes requires adjustments in several places of the code. Now, the test suite adds another layer that we must keep up to date if me make any changes to the config.ini. Wouldn't it be better to use `App.default_settings()` to create a new config on the fly during the tests? This way, we only need to update this function if we want to add new settings.
The same idea could be applied to the database: Why not use `App.new_project()` to create a new db with the most recent structure instead of creating all the tables manually?

VRehnberg:
@kaixxx I agree.

Didn't see the `App.default_settings` method. That should probably be used in this case, my bad.

The check for expected keys in config.ini I tried to keep as close as possible to as it was, but as is the value from it is quite low to the maintenance cost of keeping the `CONFIG_INI` constants up to date.

ccbogel:
That would be my fault, based on what  I originally did, when trying to work out how testing was done.

kaixxx:
@VRehnberg What do you think about using `App.new_project()` to create a new database with the most recent structure instead of creating all the tables manually? Again, this would make it easier to keep everything up to date.

VRehnberg:
> What do you think about using App.new_project() to create a new database with the most recent structure instead of creating all the tables manually?

@kaixxx I haven't really looked that much at the code so can't answer on particularts. But roughly it depends on the purpose of the tests.

- If you see it as an integration test then use the methods in combination as close to as intended as possible.
- If you see it as a unit test then you need two independent ways of generating the output.

So to take `test_load_config_ini` as an example:
```
    def test_load_config_ini(self):
        """ Tests that all config.ini fields are present """

        result, ai_models = App._load_config_ini(self)
        keys = result.keys()
        self.assertEqual(keys, CONFIG_INI_AI_EX4C0559.keys())
```
this looks like a unit test. You want to check `App._load_config_ini` in particular then the `CONFIG_INI` keys shouldn't come from the `App._load_config_ini`.

So in short, if `App.new_project()` uses `App._load_config_ini()` to generate the table, then the test wouldn't really test that much. If it creates the tables independently then it sounds like a good idea.

However, the config.ini in fixtures can probably be replaced by getting self.settings from whatever creates the default values of the config file (I didn't find this so I was sloppy). That said, there is some value in having some config files to run tests on, but these should be from specific releases to the regression testing/testing backwards compatibility.

PR #1060, opened 2025-05-21
Update es.po

Salomon-Mazatlan:
Spelling and syntax corrections were made.

Terms were standardized:
"Memorándum" and "notas" were unified as "Memo" (as appropriate).
"Gestionar" was changed to "Administrar".

A distinction was made between Graph and Chart (both of which translate to "gráfico" in Spanish). The term "Grafo" is proposed for Graph, as it refers to a network of codes and the visualization of their relationships.

Sending you a warm hug from Mazatlán, Sinaloa (Mexico).


ccbogel:
thank you

PR #1061, opened 2025-05-21
Update app_es.ts

Salomon-Mazatlan:
corrections were made, and changes were implemented for consistency with the "es.po" file in the use of terms.

Sending you a warm hug from Mazatlán, Sinaloa (Mexico).


ccbogel:
thank you

PR #1095, opened 2025-08-02
Fix flaky edit mode toggling

konbraphat51:
## Problem occurred
The edit mode toggling in the text coding screen was flaky via E+ctrl shortcut.

This is observable by keeping pushing E+ctrl.

This was because the detection method was sensing repeated key press.

## Fix
Only sense the first key press of E+ctrl


konbraphat51:
fotgot to mentions: I am Windows user

PR #1138, opened 2025-12-02
Update app_fr.ts

qwertygc:
Fix translation


ccbogel:
lets see after this is updated

PR #1141, opened 2025-12-05
Added RTF import

kaixxx:
Before, Rich Text Format (.rtf) files were imported with all the encoding markup, which made them unreadable. Now, the plain text is extracted, just like with html or docx-files. 

I am using the library striprtf for that: https://github.com/joshy/striprtf

Colin, please take a quick look at my changes if everything looks fine. I'm not so familiar with the manage files dialog. Thanks!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,3 +29,4 @@ pyinstaller<=6.4.0
 # The bootloader in the most recent version of pyinstaller is often falsely detected as malware
 # by anti virus software, so, it's safer to use a version that has been released a few months ago.
 pefile<2024.8.26  # see: https://github.com/pyinstaller/pyinstaller/issues/8832#issuecomment-2413203254
+striprtf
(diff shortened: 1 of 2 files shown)


ccbogel:
@kaixxx Yes this looks fine to me. The only concern is the character encoding used. as not all rtf documents will be latin-1.
So I might add a try except around the import

kaixxx:
From what I understand, the actual character encoding is declared inside the rtf file with a special tag. Striprtf should be able to detect that and to interpret the characters accordingly. 
But you are right, adding a try ... except for extra safety might be good.

ccbogel:
I am about to upload an update

[cid:381df966-3917-47cf-a030-4998e2909f99]
________________________________

ccbogel:
OK I downloaded about 10 rtf files including in other languages, all imported ok
________________________________

PR #1157, opened 2025-12-10
Implement export to IRaMuTeQ format

qwertygc:
Added functionality to export reports in IRaMuTeQ format.
Iramuteq is a well-known free software program for lexical analysis, particularly used in France. This feature allows you to export in a format that can be used in this software, enabling more in-depth analysis after running through QualCoder.

PS: I'm not yet very familiar with contributing to GitHub, so I apologise that this has to be done in several pull requests.


ccbogel:
OK thank you.
Itramuteq not a familiar program where  I am
I will test it out later.

PR #1161, opened 2025-12-11
Add mistral AI model

qwertygc:
Add mistal AI model, used in France and in Europe generally


kaixxx:
Great, that's nice to hear. Thanks!

PR #1162, opened 2025-12-15
Reorganisation/Cleaning of directories

qwertygc:
I have created a single file for development operations, with options:
- python devtools.py --rebuil-ui as rebuild_ui.py
- python devtools.py --update-translation as updates_po_files.py
- python devtools.py --compilation-translation as recompile_translation_po.py

Cleaning language file in the wrong directory
Various documentation files placed in the doc/ folder for clarity.

### DIFF
--- a/rebuild_lang.py
+++ b/rebuild_lang.py
@@ -90,8 +90,9 @@ def update_translation():
 
 def recompile_translation():
     project_root = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


ccbogel:
@qwertygc HI Justin, can you also chat more about changes planned, before changes are made. It becomes worrying if a lot of changes are implemented at once without discissions.
Also the .ts files there were in the wrong directory, were they placed into the correct directory.

PR #1176, opened 2026-01-05
Link to the website (help)

qwertygc:
Hello
Here is the integration of multilingual help. It is within a specific function, to make it more modular.

I took the opportunity to update the headers to link to the documentation + corrected the minor change in the French translation.


ccbogel:
@kaixxx "Add a fallback: if the French version fails to open, show the English version instead."  
No, a fallback like you state will not work with the webbrowser module. As it does not detect 404 errors.
We have to check using requests first.

PR #1187, opened 2026-01-15
fix fr translation

qwertygc:


qwertygc:
I ok purpose only french sorry I have used the script lang

ccbogel:
@qwertygc OK, can you upload only the French file that is changed

PR #1191, opened 2026-01-16
fix fr

qwertygc:
OK fix fr with only change in French


ccbogel:
thank you - there was a clash when Lorenzo also forwarded his pull request. This is good - no conflicts

PR #1232, opened 2026-03-30
Improved propagation of database changes throughout the app

kaixxx:
### The Issue:

The existing `update_dialog_codes_and_categories()` method in the coding dialogs is an attempt to propagate changes made in one dialog throughout the app so that other dialogs can update as well, for example an open report. In its current form, however, this mechanism is quite limited:

- It updates only `DialogReportCodes`, `ReportCoderComparisons`, and `ReportCodeFrequencies`, but not the other report types that have been introduced since then.
- Many operations do not trigger an update, even though they may affect open reports. For example, new codings would affect the table of code co-occurrences and many other report types.
- Finally: I also need a reliable UI update mechanism for the AI feature extensions I am currently working on.

### My Proposal:

To address these issues, I suggest a more centralized approach based on PyQt signals:

a) A simple event bus in the QualCoder app object issues a custom `project_data_changed` signal:

```python
class ProjectEventBus(QtCore.QObject):
    """Application-wide event bus for project database changes.

    This is used to notify other dialogs (e.g. reports) of changes to the
    project database, so they can update their UI (e.g. the code tree).
    """

    project_data_changed = QtCore.pyqtSignal(list, object)

    def emit_table_changes(self, tables: list[str], source=None):
        """Emit one project-change event for changed database tables.

        Args:
            tables: List of database table names that changed. An empty list
                means that no project-wide event is emitted.
            source: Optional object identifying the emitter. Subscribers can
                compare this to themselves to ignore events originating from
                the same dialog instance.
        """

        if len(tables) == 0:
            return
        self.project_data_changed.emit(tables, source)
```

b) Each report dialog subscribes to this signal and updates its UI accordingly. For example, in `report_codes.py`:

```python
def _on_project_data_changed(self, tables, source):
    """Refresh the local code tree when project events change the code system."""

    if source is self or not isinstance(tables, list):
        return
    tables = set(tables)
    if "code_cat" not in tables and "code_name" not in tables:
        return
    self.code_names, self.categories = self.app.get_codes_categories()
    self.fill_tree()
```

For now, I have implemented this for all existing `update_dialog_codes_and_categories()` calls in the coding dialogs, which covers changes to the code system.

I have also added subscriber functions (`_on_project_data_changed()`) to the relevant dialogs. However, I left out the graph report and the code organizer dialog, because I am not yet sure whether, or how, those graphs should update when, for example, a code is deleted in a coding dialog.

A next step would be to emit the `project_data_changed` signal for changes to the `code_text`, `code_image`, and `code_av` tables as well. Before implementing that, though, I wanted to get your feedback on this approach.

Don't be overwhelmed by the fact that this PR touches so many files. The changes follow a very similar pattern in all the dialogs.

### DIFF
--- a/src/qualcoder/__main__.py
+++ b/src/qualcoder/__main__.py
@@ -171,6 +171,46 @@ class ProjectLockHeartbeatWorker(QtCore.QObject):
         self.finished.emit()
 
 
+class ProjectEventBus(QtCore.QObject):
+    """Application-wide event bus for project database changes."""
+
+    project_data_changed = QtCore.pyqtSignal(object)
+
+    def emit_table_changes(self, tables, source=""):
+        """Emit one normalized project-change event."""
+
+        if not isinstance(tables, dict) or len(tables) == 0:
+            return
+        payload = {
+            "source": str(source if source is not None else ""),
+            "tables": {},
+        }
+        for table_name, table_change in tables.items():
+            if not isinstance(table_name, str) or table_name.strip() == "":
+                continue
+            change = table_change if isinstance(table_change, dict) else {}
+            ops = change.get("ops", [])
+            if isinstance(ops, str):
+                ops = [ops]
+            elif not isinstance(ops, (list, tuple, set)):
+                ops = []
+            normalized = {
+                "ops": sorted({str(op).strip() for op in ops if str(op).strip() != ""}),
+            }
+            for key in ("affected_ids", "affected_file_ids", "affected_code_ids", "affected_cat_ids", "changed_columns"):
+                values = change.get(key, [])
+                if isinstance(values, (list, tuple, set)):
+                    normalized[key] = sorted({value for value in values if value is not None})
+                elif values is None:
+                    normalized[key] = []
+                else:
+                    normalized[key] = [values]
+            payload["tables"][table_name] = normalized
+        if len(payload["tables"]) == 0:
+            return
+        self.project_data_changed.emit(payload)
+
+
 class App(object):
     """ General methods for loading settings and recent project stored in .qualcoder folder.
     Savable settings does not contain project name, project path or db connection.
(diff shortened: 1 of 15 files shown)


Salomon-Mazatlan:
Hi @kaixxx 

for: 

> I left out the graph report and the code organizer dialog, because I am not yet sure whether, or how, those graphs should update when, for example, a code is deleted in a coding dialog.

I had already implemented a feature to refresh the graph view using the “Refresh view” option; however, after reviewing your proposal, I have also integrated it into the view_graph.py module.

Preview:

update

I still need to refine a few details, but overall it is already working correctly when modifying the hierarchy, renaming items, merging codes, and updating coded segments.

kaixxx:
> after reviewing your proposal, I have also integrated it into the view_graph.py module.

Looks great, thank you.

Salomon-Mazatlan:
Kind regards, @kaixxx 


I have been working on the live graph synchronization feature, and I noticed that the coding modules do not emit change signals related to coded segments and memos. Because of this, these changes were not being immediately received by the graph module.

As a provisional solution, I added the instruction `self.app.project_events.emit_table_changes` to every method involved in coding operations. With this, I was able to restore the reception of change events in `view_graph.py`.

I am attaching a table with the methods where I implemented this modification:

[emit table changes.xlsx](https://github.com/user-attachments/files/27469986/emit.table.changes.xlsx)

My main question is whether this approach is appropriate, or if there is a better way to handle it.

kaixxx:
@Salomon-Mazatlan Yes, that looks reasonable, thank you very much for the work! 
Unfortunately, SQLite in Python does not have an `on_commit` event or something similar to automate the process. So we have to call `self.app.project_events.emit_table_changes()` basically after every commit.

ccbogel:
Just eeading through it. that's great

PR #1247, opened 2026-04-12
fix fr translate + translation information.py

qwertygc:
I fix translation and purpose a translation of information.py. But, for the last point, it's don't work exactly (it's detected on rebuild_lang.py code, but not in the software).


ccbogel:
Hello, Information.py the information there is hard-coded in the software.
If you can provide me with the translation,  I can hard-code it back into the information.py file.
Email me the translation.

PR #1259, opened 2026-04-18
small fixes

kaixxx:
In my local branch, GitHub tests reported errors, so I fixed a few minor issues. 

Some of the test files are still out of date, which is totally fine for me. I have my own local test routines anyway. I will not work on this any further. But I still wanted to share the small fixes that I've made along the way.


ccbogel:
Thanks  updated.
Yes  I also dont use tests

PR #1289, opened 2026-05-22
Update view_image.py

Salomon-Mazatlan:
**Updates and fixes:**

**-draw_coded_areas():**
--setZValue(1000) on each handle, so, always clickable on top of stacked rectangles. (interactive resize).

**-update_interactive_resize():**
--clamp the cursor to the visible bounds (with W/H swap for 90/270 rotation) before recalculating the rect, eliminates the jump and restricts resizing to the edges.

**-create_code_area():**
--clamp the rectangle to the pixmap edges instead of cancelling (return), the new coding does not exceed limits and keeps the valid part.

**-scene_context_menu():**
--menu exec() before DialogSelectItems; the selector only opens if the action depends on a segment and there are >1.  (To maintain interaction consistency, like code_text.py)  "Important" options: filtered with 1 segment, both shown with several.

**-scene_context_menu():**
--the three "Highlight this code" actions were added to segment_actions, they also disambiguate the segment (cid) when there is overlap.


ccbogel:
Thanks L

PR #1312, opened 2026-05-28
Update manage_files.py

Salomon-Mazatlan:
Some files store line breaks as \r\n (Windows style) and sometimes include an invisible character at the start (BOM). When displaying them, Qt cleans these up automatically, so the visible text ends up shorter than what was saved. 
That small mismatch makes the code positions point past the end of the text, and the highlight freezes. 
The fix cleans up those line breaks and the BOM at import time, before saving. That way the saved text and the displayed text have the same length, and positions no longer overflow. It only applies to newly imported files; files imported earlier still carry the mismatch, which is why it's worth keeping the defensive clamp in code_text as well.

### DIFF
--- a/src/qualcoder/manage_files.py
+++ b/src/qualcoder/manage_files.py
@@ -2225,6 +2225,14 @@ class DialogManageFiles(QtWidgets.QDialog):
             Message(self.app, _("Warning"),
                     _("Cannot import ") + str(import_file) + "\nPlease check if the file is empty.", "warning").exec()
             return
+        # normalise line endings and strip BOM so the stored fulltext matches
+        # exactly what QPlainTextEdit will display. Qt converts \r\n and lone \r
+        # into \n on setPlainText(), and a leftover BOM adds a char; either makes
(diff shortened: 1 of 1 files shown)


ccbogel:
Hi @Salomon-Mazatlan 
Before I pull this request. Can you test with a couple of pdf documents. to ensure that the imported text matches exactly, so that _code_pdf.py works as it is supposed to. 
I anticipate it will, but please check first.

Salomon-Mazatlan:
I’ve run some tests and haven’t found any issues, but if you’d like, we can add safeguards to prevent it from affecting .pdf files. I can make those adjustments.

Additionally, I found an error in code_pdf.py:

File "C:\Users\salom\Downloads\QualCoder-master\src\qualcoder\code_pdf.py", line 2162, in overlapping_codes_in_text
    if item['pos0'] <= pos + self.file_['start'] <= item['pos1']:
                             ~~~~~~~~~~^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable

ccbogel:
thanks if you have tested that, its great

PR #1334, opened 2026-06-03
status of translaiton

qwertygc:
It can be difficult to keep track of translations across different languages, especially when they are scattered. I have added an option (--status) that generates a statistical report on the translation. This would allow you to see how well a language is being maintained (which might encourage contributions).


ccbogel:
thanks

PR #1335, opened 2026-06-03
Update de.po

maninthemiddle01:
I edited the file in Lokalize and imported the updated and missing translations provided by Qwen 3.6 35B A3B. I hope that there aren't any incorrect characters preventing compilation now. I manually reviewed and verified all the changes, and I'm pretty impressed — I only found fewer than 10 minor mistakes. If you think this is ok, then maybe this method could be used for Simple Chinese and Japanese, because it's much better than the last translation?

### DIFF
--- a/src/qualcoder/de.po
+++ b/src/qualcoder/de.po
@@ -88,7 +88,7 @@ msgid ""
 "No codings found for this particuar combination of coder, document filter, "
 "and code."
 msgstr ""
-"Keine Kodierungen für diese Kombination aus Koder:in, Dokumentfilter und "
+"Keine Codierungen für diese Kombination aus Coder:in, Dokumentfilter und "
 "Code gefunden."
 
 #: src\qualcoder\ai_chat.py:387


maninthemiddle01 (commented):
Changed last remaining wrong Kode to Code

ccbogel:
for Chinese Japanese, what is really needed now is a human editor. the translations were made using deepl.
Is Lokalize similar to poedit?

maninthemiddle01:
Thanks for the reply! To clear up the comparison, Poedit is a tool for gettext .po/.mo files, whereas Lokalise handles over 50 formats. It also includes translation memory storage, advanced quality checks with customizable rules for consistency and terminology, and a Git workflow that auto-syncs with repos, triggers on PRs or commits, but I didn't use it.
That said, my approach was hybrid: I prompted Qwen 3.6 with a custom glossary and explicit app-context instructions, then did a manual pass. I actually tried DeepL first, but it produced some awkward strings because it translates each line in isolation without understanding the UI or how other strings relate. DeepL is fantastic for full articles, but app localization benefits from context-aware drafting followed by targeted review, that's why I thought that with my setup Qwen produced very few mistakes. If you ever want to swap notes on prompt structures or glossary setups for CN/JP, I’d be happy to compare workflows.

maninthemiddle01:
I dont know, somehow I am stuck trying to get the latest translations to show up in QualCoder.

I'm running the app on Ubuntu using the `./run_from_source_Ubuntu.sh` script. I recently updated my local files, and I can see that the new `.po` and `.qm` files for my language are definitely updated inside the folder.

However, when I launch the app, it still stubbornly shows the old translations. 

Here is what I have tried so far:
* I first installed via unpacking the zip, but because the translation didn't update (I verified with the wrong "Mmit neuem Code markieren") I removed whole directory and installed via git.
* Verified that the `.qm` files in my folder are actually the newest versions.

Is there a specific cache I need to clear, or does the `run_from_source_Ubuntu.sh` script do something with the virtual environment that holds onto the old translation files? 

Thank you much in advance for your help

qwertygc:
You have put in .qualcoder directory?

maninthemiddle01:
whole app data is in /home/user/apps/QualCoder/
The file structure inside is like the one in this repo.

ccbogel:
@maninthemiddle01 the .po files have to be turned into .mo files first. QualCoder these are the machine-readable files for the translator to use. 
And it needs to be packed in the base64_lang_helper.py file

However,  I cant look at this deeply until the weekend.
As now the .po and .ts files are re-located to another folder, and I will have to test and update the create_lang_base64.py file and also perhaps relocate the baselang64_helper.py file, and then look at removing the locale folder, and editing the main.py file to look in different locations for translations

PR #1337, opened 2026-06-04
Reorganization translation

qwertygc:
The proliferation of language files and their scattering across different folders can make external contributions difficult. Furthermore, having them mixed in with the rest of the code files reduces readability and does not encourage support for other languages.

My contribution involves placing the language source files in an “i18n” folder (short for “internationalisation” in software development). Furthermore, I have modified the rebuild_lang.py file to include, at the beginning, a single variable containing the list of supported language codes, thereby making it easier to add (or, conversely, remove) support for a particular language.


ccbogel:
ok lets try this

PR #1349, opened 2026-06-08
fix translation fr

qwertygc:


qwertygc:
I have also --update all language to have stats by language for communauty page.

ccbogel:
There are 47 conflicts to resolve. So I will go through and accept all the incoming changes?
e.g.
image

qwertygc:
Yes perfect

ccbogel:
I think I did some at 'incoming' but most as 'current'

ccbogel:
There is a problem with this request I am reverting it

PR #1353, opened 2026-06-10
Update config-ai-ec4c0559.ini

maninthemiddle01:
This integrates the GWDG SAIA API as a built-in AI provider in QualCoder. Since SAIA is fully OpenAI-compatible, it integrates into the existing AI settings. As this modifies config.ini, I want to ask if this update does overwrite or delete custom models that users may have already added to their local configurations. I have successfully tested this integration and confirmed it works.

### DIFF
--- a/tests/fixtures/config-ai-ec4c0559.ini
+++ b/tests/fixtures/config-ai-ec4c0559.ini
@@ -117,3 +117,17 @@ fast_model_context_window = 32768
 api_base = https://helmholtz-blablador.fz-juelich.de:8000/v1
 api_key = <NOT_AN_ACTUAL_KEY>
 
+[ai_model_SAIA_GWDG]
+desc = SAIA API via GWDG
+	Access to academic and research-grade models. For GWDG-hosted models,
+	requests are processed in-memory and not stored on persistent storage. 
+	To gain access and get an API-key, you must identify yourself via the GWDG 
+	institutional login. Adjust context_window according to the model.
+access_info_url = https://docs.hpc.gwdg.de/services/ai-services/saia/index.html
+large_model = alias-large
+large_model_context_window = 256000
+fast_model = alias-fast
+fast_model_context_window = 256000
+reasoning_effort = default
+api_base = https://chat-ai.academiccloud.de/v1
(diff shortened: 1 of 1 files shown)


kaixxx:
You have modified some random test file. This will have no effect on the actual application. The configs for the default models are in ai_llm.py, in get_default_ai_models(). 

A few other issues: 
- Are you sure that "alias-large" and "alias-fast" are actually the correct model names? 
- Have you checked the context window length?
- Please leave the api-key empty, so people must provide their own.

maninthemiddle01:
> You have modified some random test file. This will have no effect on the actual application. The configs for the default models are in ai_llm.py, in get_default_ai_models().

Oops, sorry about the test file! I'll move everything over to ai_llm.py.


> A few other issues:
>  * Are you sure that "alias-large" and "alias-fast" are actually the correct model names?

I thought the "alias" naming worked the same way as Blablador, but I can switch those to maybe gemma4 and qwen3.6? Or what is the best practice?

>  * Have you checked the context window length?

This is for the Gemma 4 model. Qwen 3.6 has an even bigger context length window. However, I'm also not sure what to put here because it's different for different models.

>  * Please leave the api-key empty, so people must provide their own.

Ok, thank you

kaixxx:
Please test the actual code that you are suggesting thoroughly on your local machine before creating a pull request. Do not assume that something is working, verify it. Also, try to break the code by doing silly things as normal user would do as well. 

Regarding this particular service: Please do some research first as suggested in the issue thread. If we decide to include this into the default options, it would be up to you to select a suitable model for the "large-model" and "fast-model" variants. In most cases, QualCoder uses the "large-model" only. Choose one that can also interpret subtle details in the text and is able to handle complex prompts. The fast variant can be a small model. It is only used for simple housekeeping tasks, like auto-naming an AI chat (an upcoming feature, not yet implemented in the main branch).

kaixxx:
Instructions how to access the GWDG SAIA API have been integrated into the AI Setup page: https://qualcoder.org/doc/en/2.3.-AI-Setup/ Thank you, @maninthemiddle01

PR #1364, opened 2026-06-19
Add user downloadable languages

kaixxx:
As discussed, I've updated the handling of user downloadable translations. 

- Added support for extra user translations in ~/.qualcoder/i18n.
- Added an `Add more languages...` entry to the language menu in Settings.
- Selecting that entry opens ~/.qualcoder/i18n in the system file explorer. This folder contains a README.TXT with download and file placement instructions. 
- User-provided languages in this folder are detected from compiled translation files (.qm and .mo) and added to the language selection in Settings.
- ZIP-based language packages named <lang>.zip can also be added. On startup, QualCoder extracts the contents automatically. Zip must contain <lang>.qm and <lang>.mo and may contain an optional <lang>.txt with extra information (e.g., contributor name, date, version, etc.). For security reasons, no other contents of the zip file are extracted. 
- User provided translations are now hosted separately in the "other_languages" folder in the repo. I have moved the (partial) Romanian translation over there as an example. 
- If a language exists both in the bundled application files and in the user folder, QualCoder uses the newer translation files.
- Added startup fallback to English when translation files are missing, invalid, or cannot be loaded, with a message box informing the user about the fallback.

**Extras:**
- Also added automatic language detection on first start, based on the system locale on Windows, macOS, and Linux. This way, new users can directly work in their native language (if available). 
- Added loading of Qt base translations so standard Qt dialog buttons and other built-in UI text are translated consistently.

**Open:**
- We must decide which languages will be moved to the new "user contributed" section.
- Website instructions must be updated. 
- We must decide if the `rebuild_lang.py' script should also handle language files hosted in ./other_languages. 
@qwertygc Justin, the last two questions address topics that you are most familiar with. What do you think?


ccbogel:
Thank you for this Kai

ccbogel:
I think the actively updated languages could stay as is. 
Suggest also moving it, ja, zh, and sv to the other_languages as these have not been actively reviwed by humans

qwertygc:
Thank @kaixxx for this update. For me, rebuild_lang.py must also handle language files hosted in ./other_languages. The translation workflow should be as similar as possible between officially maintained languages and those maintained by the community. This should eventually enable these languages to transition to official translation once we have found the necessary resources. rebuild_lang.py can also create the .zip

We can update the website (7.6.-How-to-contribute/ for contribution ; 2.2.-Settings for changing language).

There is one final, crucial point to address: are we agreed that the language source files, even those containing community translations, should remain in QualCoder’s main source code?

qwertygc:
One solution – the simplest from a technical point of view for `rebuild_lang.py` – is to separate the ‘source’ language files from the compiled version, for both the community and non-community versions.

Perhaps a /translations folder at the root containing the source files (ts, po), whether community or non-community. And the ‘compiled’ versions in i18n and other_languages?

I’m not sure which is best.

kaixxx:
> There is one final, crucial point to address: are we agreed that the language source files, even those containing community translations, should remain in QualCoder’s main source code?

Do you mean moving the "other languages" to a separate repo? This would be possible, but is not necessary. As it is now,
- the `other_languages` folder is not included in the compiled binaries,
- languages in there are not appearing in the settings dialog (unless the user installs them in the user directory).

So `other_languages` is really only a storage folder in the repo.

**For `rebuild_language.py`:**
I think one approach could be that the script just runs from the current directory instead of from the project root. So, if you start it from the `./src/qualcoder/i18n`, it would work with the officially maintained language files, and if you start it from the `other_languages` folder, it would work with them. We could add an extra step to the compile process, asking the user if they want to create a zip or not.
Besides, separating the source files from the compiled ones would help to clean things up, yes. We could even put the zip files in a separate folder. I would, however, not mix up official and user contributed languages, but create subdirs in the respective folders. 

What do you think?

ccbogel:
yes that sounds good

kaixxx:
@qwertygc has adapted rebuild_languages.py already: https://github.com/ccbogel/QualCoder/pull/1366

PR #1374, opened 2026-06-24
fix fr translation

qwertygc:


kaixxx:
You are so quick in translating everything to French... Thanks, Justin!

qwertygc:
I have used deepl and proofreading. There might be some tweaks to the terminology; I’ll see how it goes as I use the development version ;-)

PR #1378, opened 2026-06-25
Fix new text file dialog widget reference

kaixxx:
This was left over from the switch to QPlainTextEdit. It generates an error when creating a new blank text file.


ccbogel:
Danke

PR #1382, opened 2026-07-04
Mark Speakers

Salomon-Mazatlan:
Updates:

Captura de pantalla 2026-07-03 000348

**Speaker identifier selection.** A new "Identifier" combo box lets the user choose the marker format for the scan:

-  `Name:`, 
- `#Name:`, 
- `@Name:`, 
- `[Name]`, 
- `{Name}`, 
- a custom regular expression, or Auto-detect. 

A single format applies to the whole scan; lines in any other format are treated as continuation lines. Previously the three hardcoded formats (`Name:`, `[Name]`, `{Name}`) were all matched at once.

**Auto-detection.** Auto-detect tries each concrete format across the selected files and picks the one with the most turns. Specific formats (`#`, `@`, brackets, braces) are evaluated before the generic `Name:` so the symbol is not absorbed into the name.

**Custom regex mode.** A line edit (visible only in Custom mode) accepts a user regex.

**Multi-file support.** A "Select files" button lets the user scan several text files at once. Each coded segment keeps its own file id, so codings are written to the correct source. 

**Code with or without the speaker label.** A new checkbox lets the user code either the whole turn (including the `Name:` marker) or only the response text.

> Note: Turns with an empty response are skipped in response-only mode. Table examples now show the response text.

**Memos.** New speaker codes get a `File: <name> / Turns: <n>` memo block per file instead of the bare filename. When a code already exists, only files not yet listed are appended, so repeated runs no longer duplicate memo lines.

Captura de pantalla 2026-07-03 000415


qwertygc:
How to detect speaker when we have also timestamp. E.g : [hh:mm:ss][speaker] lorem ipsum ?

Salomon-Mazatlan:
Hi @qwertygc:

With the current formats, the module takes the first bracketed text at the start of the line as the speaker name. In [hh:mm:ss][speaker], that means it would detect the timestamp instead of the name.
___
I hadn't seen this specific format before.

Could you share which app or script produced that format?

If it turns out to be a frequent convention, I could add a predefined "timestamp bracket" format: on [timestamp][Name] lines, it would skip the first bracket whenever its content is only digits and colons (a time) and take the second one as the speaker name. Being predefined, it would also participate in auto-detection.

A more general option could also be considered: stripping leading timestamps before detection, which would additionally cover files like [hh:mm:ss] Name:.

qwertygc:
It’s a personal format that I use when I’m transcribing.

kaixxx:
That looks very well thought through. 
One question: In the "manage files" dialog, you can select multiple files, then right click and select "Mark speakers". Will these multiple files then be preselected in the mark speaker dialog? I haven't tested the code...

ccbogel:
thanks Lorenzo,

Salomon-Mazatlan:
Hi @qwertygc, In an upcoming update I could add the `timestamp_bracket` format. I am also looking into a way to support coding based on end markers.

Hi @kaixxx: for now, multiple files can be selected directly from the Mark Speakers interface; the Manage Files and Code Text modules did not require any changes.

PR #1385, opened 2026-07-07
Code organiser assist

Salomon-Mazatlan:
For new module: Code organiser assist

Manual organisation of the code system.

image

image

image

image


kaixxx:
This looks as if it could be helpful in very special situations. But it is also a bit limited for general use I would say. 
For instance: Why splitting a code into exactly two subcodes? You might want to have three, or four, or one sub-category and two subcodes, or you might want to leave some codings in the general code and move only some to the new subcodes... It all depends on the empirical phenomena under study. I think that real-world use cases will have so many different variants that it is impossible to cover them all in one "Split code" dialog, while still keeping the dialog intuitive an easy to use. 

A tree view that shows categories, codes, and codings as well, allowing for easy drag & drop, code creation and merging, might be a more intuitive user interface. But then, we are close to the already existing "code organizer" that you have improved so much over the last months. 

My gut feeling tells me it might be better to invest a little more time in making the existing code organizer perfect, smooth and intuitive, instead of creating yet another user interface that only works for very special use cases. But yeah, these are just some first impressions, based on you screenshots and graphics...

Salomon-Mazatlan:
Hi @kaixxx 

The main purpose of this module is to allow multiple changes to be made simultaneously across the code system. At present this is a limitation both in the graphical Code Organiser module and in the codes tree of the coding modules, where changes have to be made one at a time; the graphical interface also does not support filtering.

That said, in code_organiser we could implement something similar to the relationship connector I showed you in the view_graph module, in order to make the hierarchical organisation of the codes easier (I will note this idea down).

More specifically, this module could be conceived as a code manager (manage codes), offering the same reorganising, splitting, merging, editing and querying functions, along with the ability to jump directly to the files that contain the selected coding, generate specific graphs from the selection, and so on. This would broaden the functionality currently provided by the codes tree in the coding modules.

kaixxx:
"Manage Codes" would definitely be a better name compared to "Code organiser assist" :)

However, I'm still not completely convinced by the user interface, and if we really need this dialog at all. 
A tree view with drag & drop would seem more intuitive to me, with the possibility to select multiple items at once for bulk changes. Maybe two tree views side-by-side would be even better, so you can scroll to one branch on the left, another on the right, and drag & drop between them. I'm thinking about how file explorers work, because that's what people are already familiar with.

But I also have a general hesitation introducing yet another dialog or workspace. This is based on the fact that many people criticize tools like NVivo for being bloated with features and too complicated to use. QualCoder, on the other hand, is praised for its simple interface. We should be careful to not put this advantage at risk, even if a small number of power-users might welcome such a new feature. 

BTW, because I was talking about making the code organiser "perfect, smooth and intuitive", here are some minor issues I've found while testing: https://github.com/ccbogel/QualCoder/issues/1386

kaixxx:
Wait, you have closed this - that wasn't my intention. Let's continue the discussion, also with @ccbogel!

Salomon-Mazatlan:
Hi @ccbogel 

I think I was too hasty, probably because I was tired.

I am going to make a few adjustments and present the proposal in the forum section instead.

@kaixxx, I closed the pull request so I can present the proposal in a more appropriate way.
The idea is to do something like this: (concept art)

image

image

PR #1390, opened 2026-07-08
Code organiser update

Salomon-Mazatlan:
**Updates**

**Sub-codes:** 
nest a code under another with recursive depth, catid/supercid mutual exclusivity, id-based cycle prevention, and merging a parent code carries its sub-codes to the target.

**Model-first editing with a transactional Apply:**
everything lives in the model (temporary ids) until Apply, which shows a change preview, runs as a single transaction (full rollback on any failure), writes to the journal, and notifies the event bus.

**Codebook import:** 
QDC (REFI-QDA), CSV, and TXT, with a reuse or import-as-copy policy.

**Tree-based branch selector:** 
the full category/code/sub-code tree (any node or All), with code colours, a reusable button, and no detaching of the root on Apply.

**Hierarchical connect handle:** 
drag one node onto another to nest it, with cycle and invalid-relation guards (Esc cancels).

**Canvas usability:** 
parent-to-child arrowheads, lines visible in dark mode, fit and center (menu and loupe), smoother zoom/pan, and the canvas no longer follows the node while dragging.

**Coding-frequency** 
[n] labels per code, optional.

image

image

image


kaixxx:
Hi Lorenzo, I have tested the changes, and this is fixes almost all my issues, great. 
A few notes: 
- The loupe button is now clickable and works well. However, it is not styled like the other buttons (missing border).
- I really like the new (or just bigger?) arrows, as the make it much clearer which element is the parent and which is the child. However, I find it a bit unintuitive that you must drag a line from the *child* to the *parent* to reorganize the relationship. I would expect it to be the other way round, following the direction of the arrows. Also, in the real world, it's usually the parents who decide to have children, not the other way round... ;)
- If I press SPACE, the mouse pointer changes to a hand. But I can still not drag the viewport around - at least I didn't understand how. This would be convenient, but you could also decide to remove it altogether if it is too hard to implement.

kaixxx:
One last thing to add:
- If you have unsaved changes in the code organizer, but switch to a different coding dialog or close the app, it would be good to have a message asking if you want to apply the changes first, discard them, or cancel the swtiching/colsing operation.

Salomon-Mazatlan:
Hi @kaixxx 

- I'll take a look at the loupe button, I really didn't pay much attention to it.
- The logic I had in mind for the drag handle for hierarchical positioning is one of inclusion, like in the code tree where you drag a category or code and nest it into the target item.
- I still need to implement the ability to select and nest several items simultaneously, but maybe I'll work on that in a future update.
- I limited the panning on the canvas to the area covered by the items or nodes. I know it's a bit tedious; I haven't been able to make it more fluid. It worked for me in view_graph, but not here, I'm overlooking something.
- I also left out the ability to add different code branches: every time one is loaded, the viewer resets. I implemented it that way to avoid breaking the cycles in the subcodes system.
- By the way, if you get a chance to test the codebook import, I haven't been able to test it with .qdc files from NVivo; some that I downloaded from the internet failed to import.

Salomon-Mazatlan:
> One last thing to add:
> 
> * If you have unsaved changes in the code organizer, but switch to a different coding dialog or close the app, it would be good to have a message asking if you want to apply the changes first, discard them, or cancel the swtiching/colsing operation.

That's a good idea. I had been thinking about looking for a way to save the state of the organiser, for example, if I go to code text and come back to the organiser, being able to continue the work, or something like how graphs are saved.

kaixxx:
> The logic I had in mind for the drag handle for hierarchical positioning is one of inclusion, like in the code tree where you drag a category or code and nest it into the target item.

Ah, I understand. However, this is still going against the direction of the arrows. A neat solution could be to allow dragging in both directions. What do you think?

> I still need to implement the ability to select and nest several items simultaneously, but maybe I'll work on that in a future update.

Yes, leave that for a future update. It would also require a way to visually mark selected items, maybe with a thicker border.

> I limited the panning on the canvas to the area covered by the items or nodes. I know it's a bit tedious; I haven't been able to make it more fluid. It worked for me in view_graph, but not here, I'm overlooking something.

It's totally fine to limit the panning to the area covered by the nodes. My issue is that I cannot pan at all. Maybe I'm missing something. If I press space, then click and drag around with my mouse/trackpad, nothing happens, even if I am fully zoomed in on a single node...

> I also left out the ability to add different code branches: every time one is loaded, the viewer resets. I implemented it that way to avoid breaking the cycles in the subcodes system.

Yes, that's fine, maybe something for a later update. Today, you can still decide to add all codes at once, right?

> By the way, if you get a chance to test the codebook import, I haven't been able to test it with .qdc files from NVivo; some that I downloaded from the internet failed to import.

Unfortunately, I don't have access to NVivo. But I think @ccbogel has. Maybe he can provide you with an example file.

Salomon-Mazatlan:
> Ah, I understand. However, this is still going against the direction of the arrows. A neat solution could be to allow dragging in both directions. What do you think?

I like the idea, I'll do some tests.

> Yes, that's fine, maybe something for a later update. Today, you can still decide to add all codes at once, right?

Yes, they can all be added in ALL

PR #1392, opened 2026-07-08
Update speakers.py

Salomon-Mazatlan:
**Updates:**

image

image

- Removed Auto-detect; the control is now a multi-select checklist: `Name:`, `#Name:`, `@Name:`, `[Name]`, `{Name}`, `Custom (regex)`, with `Name: ` first and checked by default.

- Multiple identifiers can be combined.

- Custom (regex) can be combined with the fixed formats.

- Detection rules based on what is checked: a single fixed format, anchored at the start of the line; regex alone, anywhere in the line; two or more (mixed mode), the symbol markers and the regex anywhere in the line, with several per line and the text before the first marker staying with the previous turn.

-- Clicking the name leaves only that identifier (direct switch, closes the dropdown).
-- Clicking the checkbox adds or removes it from the combination (without closing the dropdown).

- Regex anywhere in the line. The custom regex now matches the marker anywhere in the line (not just at the start), with support for several markers per line.

- `~\w+:` works without a capturing group.

- Blank lines no longer end the turn; a turn spans several paragraphs until the next marker or the end of the file, without including the trailing blank lines.

- Multiple codes with ; in the "code as" column: the same turns are coded to each code.

image

- Segment preview via right-click on a row: a table with a checkbox per segment (file, position, excerpt) to select which segments to code.

image

image

- Mid-paragraph Name: detection
Name: is detected at the start of the line (the ideal case) and also within paragraphs, under strict conditions: an uppercase name glued to the colon (Salomon: or El salomon:), and if there is a period or comma before it, the name is the words between the separator and the colon.

- "Additional codes" column
New column after "code as": an editable combo per row where you type new codes (several with ;) or pick from a dropdown list that accumulates the ones already used in other rows; picking one adds it to that cell's codes. 

- They are applied to the speaker's turns in addition to the main code, the vocabulary is shared across rows and persists across re-scans.

image

- Safeguard against accidental application
Enter/Return no longer closes or applies (OK only via manual click). Escape still closes without applying.

Possible improvements:
- Auto-detect additional codes (for example, codes used as attributes) from the identifiers.
- Include filters or exception lists (maybe)

### DIFF
--- a/src/qualcoder/speakers.py
+++ b/src/qualcoder/speakers.py
@@ -291,11 +291,11 @@ class DialogSpeakers(QtWidgets.QDialog):
         # El orden pone Name: primero, luego los formatos con simbolo y al final Custom.
         # Order puts Name: first, then the symbol formats, and Custom last.
         self._identifiers: List[Tuple[str, str]] = [
-            ('name',    'Name :'),
-            ('hash',    '# Name :'),
-            ('at',      '@ Name :'),
-            ('bracket', '[ Name ]'),
-            ('brace',   '{ Name }'),
+            ('name',    'Name:'),
+            ('hash',    '#Name:'),
+            ('at',      '@Name:'),
+            ('bracket', '[Name]'),
+            ('brace',   '{Name}'),
             ('custom',  _('Custom (regex)')),
         ]
         combo = self.ui.comboBox_identifier
--- a/src/qualcoder/speakers.py
+++ b/src/qualcoder/speakers.py
@@ -24,76 +24,149 @@ import logging
 import re
 import datetime
 from typing import Any, Dict, List, Tuple, Optional
-from PyQt6 import QtCore, QtWidgets
-import qtawesome as qta  # <- L
+from PyQt6 import QtCore, QtGui, QtWidgets
+import qtawesome as qta
 from random import randint
 import sqlite3
 
 from .GUI.ui_dialog_speakers import Ui_Dialog_speakers
 from .color_selector import colors, colour_ranges
 from .helpers import Message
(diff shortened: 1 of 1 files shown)


kaixxx:
I haven't tested the code, but judging by the screenshots and description, this looks nice. 
A few notes: 
- Maybe the "Mid-paragraph Name detection" should be optional, with a tick box. I could imagine situations in which this mid-paragraph detection could lead to issues.
- In your screenshot, it looks as if there are spaces in the patterns like "# Name :", and "[ Name ]"? Or am I seeing this wrong? 
  grafik

Salomon-Mazatlan:
The only issue with mid-paragraph detection occurs when using the "Name:" identifier, although an option to disable it could certainly be added.

I implemented it this way because several transcription applications do not include line breaks in the generated files.

Yes, those are spaces. I forgot to remove them, although they do not affect the detection. Originally, I used placeholders such as "@+Name+:" simply as visual guides to make the patterns easier to distinguish.

kaixxx:
If you want to make changes to this PR (e.g., remove the spaces in the dropdown), you can simply change that in your `patch-1` branch, and your new commits will be added here automatically. Just in case you didn't know...

Salomon-Mazatlan:
> Just in case you didn't know...

hehehhe yes, I get pretty lost among the options on this site

Salomon-Mazatlan:
> If you want to make changes to this PR (e.g., remove the spaces in the dropdown), you can simply change that in your patch-1 branch, and your new commits will be added here automatically. Just in case you didn't know...

I already made the adjustment

PR #1397, opened 2026-07-09
Minor updates

kaixxx:
- In https://github.com/ccbogel/QualCoder/commit/3859e0e5b6015718676ff26be2e136ff3bd84457, some of the AI related menu updates in __main__.py where deleted, e.g., graying out the AI related menu entries if the AI is disabled. I guess this happened accidentally, so I've restored the changes. 
- Besides, I've made a small update to the way the default project folder is determined. I've learned that Qt has a nice function for that (QtCore.QStandardPaths), so I've created a small helper function around that. This also ensures that the folder exists.


ccbogel:
Danke

PR #1405, opened 2026-07-16
Readme linux

qwertygc:
Hello,

There is a great deal of overlap between the different versions of Linux, with only a few changes required when installing dependencies. Here is a proposal to simplify all of this.


ccbogel:
Trex bien . good idea.
I had a chance to test Qualcoder on Fedora 44 today.

PR #1408, opened 2026-07-19
Some code stripes refinements

kaixxx:
I have worked a little bit on refining the code stripes view in code text, which I like a lot and use as my default view. 
Please take a look, @Salomon-Mazatlan - are you OK with that?

**Readability of code labels**
One issue was that depending on the color of the code, the contrast of the code label and the background was too low, leading to poor readability.
- I changed the code margin background to the same color as the main text edit and the line number bar, which already enhances contrast a lot. 
- To further improve visibility, I defined a minimum contrast between background and code color. In a dark theme, very dark code colors are lightened to bring the contrast above the minimum level. In a light theme, light code colors are darkened respectably. This only applies to the label and the underline, not the code stripe. 
- In the example below, the dark red color of "Work..." is lightened slightly, all other colors stay the same. 

grafik

<br />

**Code stripes margin width**
The selected width of the margin with the code stripes was not persisted in config.ini. Also, the minimum width was a little wide, especially when showing the new AI agent sidebar on the right. 
- Default width is still 100px, but minimum was reduced to 30px. On widths below 60px, only the stripes are shown, not the labels. 
- Width is now saved to/restored from config.ini

**Minor updates**
- Mouse-wheel scrolling now also works when hovering the coding margin or line-number area
- Tooltip flicker for code stripes fixed
- Minor changes to some AI features are also included in this PR. I've fixed a few problems along the way. Sorry for not keeping them separated.


Salomon-Mazatlan:
Hi @kaixxx,

This looks great! I was actually just looking for a way to improve code visibility when using the software's dark theme.

 I'll apply these tweaks to the code_pdf module proposal:
https://github.com/ccbogel/QualCoder/compare/master...Salomon-Mazatlan:QualCoder:Cosas-de-pdf-2

ccbogel:
Thanks Kai

PR #1411, opened 2026-07-20
Update code_organiser.py

Salomon-Mazatlan:
New "Delete code" and "Delete category" actions in the node's context menu (code_organiser.py):

image

- With prior confirmation. They only modify the in-memory model; the database is written when Apply is pressed, within the existing single transaction (nothing is left half-applied).

- Deleting a code also removes all its codings (code_text, code_av, code_image), just like deletion from the code tree. Its sub-codes are reparented (they move up to the grandparent code, or move to the category / top level).

- Deleting a category leaves its codes and sub-categories at the top level.

- New items (not yet saved to the DB) are simply removed from the model.

- The change preview distinguishes "Code deleted (all its codings will be deleted)" and "Category deleted" from a "merged".

- Items marked for deletion are excluded from the merge/link lists, from the scene cleanup, and from the links drawn on the canvas.

### DIFF
--- a/src/qualcoder/code_organiser.py
+++ b/src/qualcoder/code_organiser.py
@@ -36,6 +36,7 @@ from PyQt6.QtWidgets import QDialog
 from .add_item_name import DialogAddItemName
 from .code_in_all_files import DialogCodeInAllFiles
 from .color_selector import TextColor, colors as valid_colors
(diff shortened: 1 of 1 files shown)


ccbogel:
thank you. I tried looking at the code organiser code for this and found i a bit hard to follow

PR #1412, opened 2026-07-20
Mark speakers -> preselección de archivos

Salomon-Mazatlan:
Preselection rules when opening the dialog:

From Manage files: all selected and visible rows are carried over (not just the last one), passing the complete list through the new doc_ids parameter.

If there is no usable selection (nothing selected, or the selection was hidden by a filter): the preselection is all visible files, so an active filter carries over. The dialog can no longer be opened "with no files".

From Code text: the open file, or all visible files if none is loaded.

When opening Select files, the current selection appears marked so it can be reviewed and adjusted (select_items.py, new preselected parameter).

New label next to the button showing the selected files with an ellipsis and a tooltip with the complete list.

image

New checkboxes: multiple files can be selected by ticking the check or with Ctrl/Shift+click; the result is the union of both. In checkbox mode the preselection is ticked (without phantom highlighting), so unticking does deselect.

image


kaixxx:
Works very well, thank you! 

The only thing I find a bit confusing is that both the checkboxes and normal highlighting is somehow "merged" into a joint selection in the "select files" dialog. If I see checkboxes, I would expect them to indicate the current selection, and I would be very surprised if a file that has no check mark would still end up in my selection... I would change this.

Salomon-Mazatlan:
> Works very well, thank you!
> 
> The only thing I find a bit confusing is that both the checkboxes and normal highlighting is somehow "merged" into a joint selection in the "select files" dialog. If I see checkboxes, I would expect them to indicate the current selection, and I would be very surprised if a file that has no check mark would still end up in my selection... I would change this.

@kaixxx ,You're right, I'll look into it.

PR #1414, opened 2026-07-22
Update code_organiser.py

Salomon-Mazatlan:
**Fix for "Spacebar + drag does not work on my system" (Kai's observation)**

-- The root cause was a focus issue: space-bar panning relied on keyPressEvent/keyReleaseEvent at the QDialog level, but those key events only reach the dialog if it has focus. In practice, when focus was held by a button (Apply, Export, etc.) or by a graph node, the space key never reached the handler that sets _space_pressed, so the subsequent drag did not pan (and worse: if a button had focus, space would "click" it).

**Context menu for multiple selection**

image

-- When more than one node is selected (rubber-band or Ctrl+click) and you right-click on one of them, a group menu now appears instead of the individual menu.
-- It detects whether the selection contains only codes, only categories, or is mixed, and shows a header with the count, with each action operating on its own type:
--- With codes in the selection: Link codes to category, Link codes under code, Delete codes.
--- With categories: Link categories under category, Delete categories.
--- In a mixed selection, both groups appear as separate actions.

**Note:**
-- Excluded: merges, "add code/sub-code/category", "Display memo", "Coded/Case text and media", rename and memo: those remain individual-only.

### DIFF
--- a/src/qualcoder/code_organiser.py
+++ b/src/qualcoder/code_organiser.py
@@ -111,8 +111,11 @@ class CodeOrganiser(QDialog):
         self.ui.graphicsView.setDragMode(QtWidgets.QGraphicsView.DragMode.RubberBandDrag)
         self._space_pressed = False
         self._is_panning = False  # middle-button / space+drag panning
-        # hierarchical drag-to-connect handle (same UX as the relations
-        # handle in view_graph). Click a selected node's blue handle, move, then click
+        self.ui.graphicsView.setFocusPolicy(QtCore.Qt.FocusPolicy.StrongFocus)
+        self.ui.graphicsView.installEventFilter(self)
+        self._initial_canvas_focus_done = False
+        # hierarchical drag-to-connect handle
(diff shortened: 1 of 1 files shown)


kaixxx:
Thank you for taking a look at this issue. Unfortunately, it is still not working for me. I get the same behavior you describe: If a random button has focus, hitting space will trigger the button. If I click the graph window first to shift the focus, hitting space and dragging does nothing. Am I doing something wrong?

Salomon-Mazatlan:
> Thank you for taking a look at this issue. Unfortunately, it is still not working for me. I get the same behavior you describe: If a random button has focus, hitting space will trigger the button. If I click the graph window first to shift the focus, hitting space and dragging does nothing. Am I doing something wrong?

I haven't been able to pin down the cause of the problem. On my computers it works correctly, but they're all Windows (10/11). I'll take another look at it, or otherwise disable this feature and perhaps use the view graph minimap instead.

PR #1415, opened 2026-07-22
Update code_text.py (Proposal with partial testing, for review.)

Salomon-Mazatlan:
**Proposal with partial testing, for review.**

tree_menu: new Delete category branch action in the Modify submenu, directly below Delete, shown only when the selected item is a category. Dispatch returns immediately, as the tree item is invalid after deletion.

get_branch_catids_and_cids(catid): resolves the branch from the database rather than from the dialog's cached self.codes / self.categories. Collects the category, all descendant categories via supercatid, all codes whose catid falls in the branch, and all nested sub-codes via the supercid chain. Iterative walk, so malformed or cyclic data cannot cause unbounded recursion.

delete_category_branch(selected): confirmation and cascading delete.

image

### DIFF
--- a/src/qualcoder/code_text.py
+++ b/src/qualcoder/code_text.py
@@ -3167,6 +3167,10 @@ class DialogCodeText(QtWidgets.QWidget):
             action_merge_category = modify_menu.addAction(_("Merge category into category"))
             action_move_category = modify_menu.addAction(_("Move category under category"))
         action_delete = modify_menu.addAction(_("Delete"))
+        action_delete_branch = None  # <- L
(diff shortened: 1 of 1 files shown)


ccbogel:
Yes sure, .you are so fast doing this If it works well to ad to the other coding pages AV, image, PDF

Salomon-Mazatlan:
> Yes sure, .you are so fast doing this If it works well to ad to the other coding pages AV, image, PDF

I had already implemented this in one of my test builds, I just ported it over with a few adjustments. Hehehe, I've been playing around with the program quite a bit, so I have a few things ready ahead of time.

PR #1419, opened 2026-07-23
Code tree controller (for review)

Salomon-Mazatlan:
code_tree.py module: brings together in a single place all the behaviour of the code tree: how it is loaded, its context menu, drag and drop for reorganising, the F2 to F6 shortcuts and category branch deletion.

This behaviour is now included in every coding module (text, image, audio/video and PDF).

Bugs fixed:

- Moving a code could create a circular reference. When moving a top level code, its own sub-codes appeared in the list of destinations. Choosing one left parent and child pointing at each other in the database, with a broken tree.
- The category frequency calculation had a badly written condition (or instead of and) in all four modules: its safety limit never took effect, so a circular reference between categories left the application hanging. On top of that, with normal data it ran 10,000 empty iterations every time the counts were refreshed.

 - - the same condition is also present in reports.py and view_charts.py

- F6 on a category in view_image called the move method with the wrong data type.

### DIFF
--- a/src/qualcoder/code_av.py
+++ b/src/qualcoder/code_av.py
@@ -17,8 +17,8 @@ If not, see <https://www.gnu.org/licenses/>.
 Author: Colin Curtain (ccbogel)
 https://github.com/ccbogel/QualCoder
 https://qualcoder.wordpress.com/
-https://qualcoder.org/
 https://qualcoder-org.github.io
+https://qualcoder.org/
 """
 
 import sqlite3
@@ -29,7 +29,6 @@ import logging
 import os
 import platform
 import qtawesome as qta  # see: https://pictogrammers.com/library/mdi/
-from random import randint
 import re
 import subprocess
 import time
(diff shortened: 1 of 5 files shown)


ccbogel:
Great work. 
Oops - Thanks for fixing my coding errors.

How do you have the time to do this so fast?

Salomon-Mazatlan:
> How do you have the time to do this so fast?

That's the positive aspect of the tumour, xD
It doesn't let me sleep, so I try to focus by working on the computer. hehehehe

Salomon-Mazatlan:
A question:

On Mac or Linux, is the "Delete" or "Backspace" key different?

 In view_graph I mapped both of them to the remove node action, but I see that F4 is used here, so if it behaves differently I'd like to use the same key in the view_graph update.

kaixxx:
This is a very useful refactoring which makes future updates much easier. Nicely done. 

I have tested it quickly, and everything seems to be working as before. 

I would use `Delete` as a shortcut for deletion, not `F4`. I have checked that it only works when the tree has focus, and we get an extra confirmation dialog, so it's pretty safe already. On macOS, I would add `Backspace` as well, since this is triggered when hitting the `Delete` key on the keyboard. Something like:
```python
shortcuts = QKeySequence.keyBindings(
    QKeySequence.StandardKey.Delete
)

if sys.platform == "darwin":
    mac_delete_key = QKeySequence(Qt.Key.Key_Backspace)

    if mac_delete_key not in shortcuts:
        shortcuts.insert(0, mac_delete_key)

delete_action.setShortcuts(shortcuts)
```

PR #1421, opened 2026-07-23
fix fr refresh other languages (computer translation)

qwertygc:


kaixxx:
Thanks for extracting the language files. But please do not do any automatic translation on the German files. I have my own pipeline for that with a local vocabulary file to keep things consistent.
 
Also: There are a lot of changes in the language files. My AI (Codex) has counted 808 changed/newly translated entries in the German .po file alone, while 539 of them are marked as "fuzzy". Both numbers seem very high. As I have updated the German translation about two or three weeks ago, I find it a bit hard to believe that so many new or modified English strings have accumulated in such a short period. Do you have an idea why these numbers are so high? As you know, checking 800 translated strings manually is a lot of work...

qwertygc:
It's the number... it's @Salomon-Mazatlan update

qwertygc:
What is your pipeline? I m interresed for enhanced own.

kaixxx:
> It's the number

I have investigated a little more:
- This PR also seems to bring back into active state many translations previously marked as obsolete (200+ entries with `#~`), e.g., "AI provider". They are not "fuzzy", however, so could be ignored in the manual check.
- Besides, some already translated strings have been retranslated and marked as "fuzzy". In some cases, I can see why: The old translation had some minor issues. But the new translations are often much worse and completely destroy the meaning. Some examples: 
```
"Hide frequencies"
Old: Code-Frequenzen
New: Hüllenfrequenzen (meaning something like "envelope frequencies")

"Show frequencies"
Old: Code-Frequenzen
New: Sendefrequenzen (meaning "broadcasting frequencies")
```
- In addition to that, there are indeed a few hundred new strings that genuinly needed translation. They are marked as "fuzzy" as well, so easy to check.

Overall, I'm not sure whether we should go forward with this PR. At least for the German version, it would probably be less work for me if I had a .po file with just empty German strings for the new entries only, no fixes, no reactivation of obsolete strings, no automatic translation. 


> What is your pipeline?

It's in a very early stage and closely tied to Codex, so not very helpful to share at the moment. 
I took inspiration from you approach and added some ideas - the main being to automatically extract and curate a vocabulary file with key concepts in QualCoder, so the translation of these concepts will be consistent throughout the app. For instance, I have decided on a canonical translation for "Journals", which was pretty mixed up before (sometimes "Tagebücher", sometimes "Notizbücher", ...). I think the fundamental approach is a real improvement, my current implementation not so much... ;)

qwertygc:
Right. I’ll create a commit with an older version of the German text later on, when I get back from my walk.

qwertygc:
OK, rollback for german language.

Maybe we need have discussion to create a thesaurus, in different langues to key concepts in QualCoder. Indeed, this would be useful even for documentation, as it would allow for a more user-friendly approach (especially if the terms differ from those used in other software).

kaixxx:
Thank you very much, Justin. I hope I didn't disturb you walk :(

Yes, I totally agree. Ensuring a consistent naming of key concepts and menu entries between the UI translations and the documentation would be very desirable as well, especially if we use some form of semi-automated process. This would reduce the manual correction by a lot. I don't know if we need a full-fledged thesaurus, but some form of shared vocabulary list would be helpful.

Also, from my experience, it really helps with semi-automated translations to add some context information about the app. This is only possible with an LLM, as far as I can see. DeepL only looks at one isolated string at a time, without any context. This is why it thinks that "code frequencies" might have something to do with radio broadcasting ("Sendefrequenzen")... If you use an LLM for translation, you can add in more context to avoid such misinterpretations.

qwertygc:
This brings me back to one of my frequent suggestions: use Weblate, which includes this feature. https://weblate.org/fr/

kaixxx:
> this brings me back to one of my frequent suggestions: use Weblate

I wasn't aware that Weblate also has LLM integration. I've found it now, looks promising: https://docs.weblate.org/en/latest/admin/machine.html This would indeed be yet another argument for using this platform. Thanks, Justin.

PR #1424, opened 2026-07-24
Cosas de pdf 3

Salomon-Mazatlan:
**code_pdf.py**
- Continuous viewer with lazy threaded rendering
- Text and area coding
- Coding margin matching code_text (WCAG contrast, tooltips, persistent width)
- Dedicated text extractor with paragraph layout and dual-variant verification
- Lines <-> paragraphs restructuring
- Highlight and annotation detection
- Autocoding
- Single page or whole document view mode (persistent)
- Forced worker shutdown on application exit

**manage_files.py**
- PDF import using the viewer's extractor (paragraph layout)
- (New) Optional highlight coding ("PDF Highlights" category, colour-family codes with suffix, comment memos, idempotent); annotations appended to the file memo
- PDF preview with "Convert to txt", editable copy extraction
- PDF to images
- Highlight progress unified into the same import dialog
- Cleanup of copies on failed imports

**code_text.py**
- PDFs remain listed and codable but text editing is blocked (points to the editable copy); reloads on external source changes; removed upstream ris_res debug prints

**main.py**
- PDFs routed to the PDF coding view
- New settings keys registered (margin width and view mode)

**helpers.py**
- Area context viewer: in-memory PDF render (no temp file, document closed, range guard), annotations not painted, pdf_page NULL -> 0 and robust access with .get()

**view_image.py**
- In-memory PDF render
- Page reset on file change, normalization of legacy areas with NULL pdf_page
- Event bus emission

**view_graph.py**
- Three in-memory PDF renders without annotations
- Both area context viewers now query and pass pdf_page

**code_in_all_files.py**
- Both PDF area renders in memory, pdf_page NULL -> 0, and source_path initialized (latent UnboundLocalError)

**report_codes.py**
- In-memory PDF area render
- PDF detection by mediapath (areas with NULL pdf_page were read as an image file)

**report_comparison_table.py**
- The image query now includes pdf_page, so the context viewer opens the correct page

**report_cooccurrence.py**
- Image coding co-occurrence:
- Areas intersecting on the same file/page, geometric E/I/O relations
- Merge creates the union area in code_image
- Refresh now watches code_image

**report_code_summary.py and report_file_summary.py**
- Fix for the category frequency loop (or -> and in the guard, prevented freezing with cycles)
- Consistent renders

**refi.py**
- REFI-QDA import/export of areas on PDF pages, with rollback on duplicates

**ai_vectorstore.py**
- "Database is locked" fix in search.sqlite: old chunk deletion moved to after the embedding, reducing the write-lock window from minutes to milliseconds; only affects the derived database, never the project database

**text_file_replacement.py**
- Blocked replacement of PDF sources in Special Functions (protects the page mapping)


ccbogel:
Gracis

ccbogel:
Lorenzo, the new version does it change the plain text importation of PDFs, compared to older versions? 
Will need to warn users of this in the documentation and release notes. Because if the older project is opened into version 4.0, they will not be able to go back to opening the project again in an older version of QualCoder.

Salomon-Mazatlan:
Plain text importation (txt, docx, odt, epub, html, md) is completely untouched, same code path as before. 

Backward compatibility is preserved in both directions: the PDF coding view verifies both text reconstructions (the classic one-line-per-visual-line layout and the new paragraph layout), so PDFs imported with older versions open and code normally, no conversion needed. 

And this PR adds no database schema changes, no new tables, columns, or migrations, so on its own it doesn't prevent a project from being reopened in an older version. 

For the documentation: PDFs imported in 4.0 store their text in paragraph layout and are coded in the new PDF view; if such a project is opened in an older QualCoder, those PDFs simply appear as regular text sources, codable in the text view as always.

QC 4.0 .............................................................................................................................................. QC 3.8.2

image

image


Note: Only the text would not match in the previous code_pdf module. But it should be correctly coded in code_text.

Salomon-Mazatlan:
**Additional note:** I didn't test on versions earlier than 3.8.x

kaixxx:
Just a quick note regarding the new PDF dialog in general: I have tested it a little, and it is really a big improvement. I haven't found any issues so far. The integration with the AI agent is also very helpful. Great that source references from the AI are now opened directly in the PDF window. This might open up a whole new field for QualCoder: Using it as a serious platform for AI assisted literature reviews. I'm curious to see what people will do with it. 
Thank you for all your hard work on this!

Salomon-Mazatlan:
> Just a quick note regarding the new PDF dialog in general: I have tested it a little, and it is really a big improvement. I haven't found any issues so far. The integration with the AI agent is also very helpful. Great that source references from the AI are now opened directly in the PDF window. This might open up a whole new field for QualCoder: Using it as a serious platform for AI assisted literature reviews. I'm curious to see what people will do with it. Thank you for all your hard work on this!

Thanks, that's exactly what I'm testing: AI-assisted literature reviews. Blablador has improved quite a bit, it now works for creating and assigning codes in files (or maybe it always worked and I just didn't know how to use it, hehe).

image

kaixxx:
> Blablador has improved quite a bit

This is mainly because the small to midsized models run by services like Blablador have improved a lot recently. Besides, the new AI agent in QualCoder can self-correct in case of small errors, which also helps a lot.

PR #1426, opened 2026-07-24
fix fr

qwertygc:


qwertygc:
+ add check code in rebuild_lang (helping translation)

PR #1429, opened 2026-07-26
Corrected a few strings in fr translation

kirmaha:
Main contribution is a correction of a string that did not mean anything in the French translation ("Assign case to file", currently "Affecter un cas à un autre" and replaced by "Assigner un cas au fichier").

Note: I intended to address another issue, but it appears it is not related to translation so I don't know how to fix it -> in the file window, when you right click and select "Assign case to file", the dialog window that opens is titled as "Delete files" ("Supprimer des fichiers"). It should rather be titled as "Select case".


ccbogel:
Merci, I have fixed this now:
In the file window, when you right click and select "Assign case to file", the dialog window that opens is titled as "Delete files" ("Supprimer des fichiers"). It should rather be titled as "Select case".

qwertygc:
I don’t know why, but there’s some sort of pre-translation that’s inconsistent. I correct it every time, but it would be nice if it stopped… I check in regularly to incorporate the new translations.

PR #1441, opened 2026-08-01
Update manage_files.py

Salomon-Mazatlan:
add: 

Zoom controls (25%–200%) to the PDF viewer and PDF-to-images converter, with sharp full-size rendering in a scroll area and re-entrancy guards to prevent resize crashes on Windows.


ccbogel:
thx for fixing

PR #1446, opened 2026-08-01
fix: bump langchain-openai minimum version to 1.3.1 to resolve startup ImportError

maninthemiddle01:
**Problem**  
Running `run_from_source_Linux.sh` fails with:
```
ModuleNotFoundError: No module named 'langchain_openai.chat_models.codex'
```
The traceback shows the import chain `__main__ → ai_chat → ai_llm` breaking at line 47 of `ai_llm.py`.

**Root Cause**  
`ai_llm.py` imports private/experimental LangChain classes (`_ChatOpenAICodex`, `_ChatGPTOAuthRefreshError`, `_FileChatGPTOAuthTokenProvider`, etc.) that were only introduced in `langchain-openai>=1.3.1`. However, `requirements.txt` currently specifies `langchain-openai>=0.1.16`. 

When pip resolves dependencies in an existing virtual environment, it satisfies the constraint with older versions (e.g., `1.2.2`), which lack these modules. Fresh clones work because pip pulls the latest release, which masks the issue during development.

**Fix**  
Update the version constraint in `requirements.txt` to guarantee the required modules are present:
```diff
-langchain-openai>=0.1.16
+langchain-openai>=1.3.1
```

**Notes**  
- The imported classes are underscore-prefixed and explicitly documented as experimental/unofficial by LangChain. They aren't exported in `__all__` and can be removed or refactored in any future release.
- Consider wrapping these imports in a `try/except ImportError` block. If they fail, the app could gracefully disable only the ChatGPT-OAuth profile instead of crashing at startup.
- Adding an upper bound (e.g., `>=1.3.1,<2.0.0`) would help prevent unexpected breakage from major LangChain updates.

**Verification**  
- Upgrading to `1.3.1+` resolves the import error and allows the application to start normally.
- Existing environments will automatically upgrade on the next `pip install -r requirements.txt`.
- No other dependency conflicts observed; `langchain-core` and related packages resolve cleanly.


ccbogel:
OK thank you.

Issue #1417, opened 2026-07-22
Feature request: Import LaTeX .tex documents via optional Pandoc conversion


stoicfist:
QualCoder currently supports formats such as DOCX, ODT, PDF, Markdown, and plain text. Researchers who write transcripts, interview material, or research documents in LaTeX currently need to convert these files manually before importing them into QualCoder.

I am using QualCoder for an interdisciplinary research project in which parts of the documentation and source material are maintained in LaTeX. For this workflow, direct .tex import would reduce manual conversion steps and make it easier to keep the original source file connected to the imported text.

I would like to implement an optional .tex importer that uses Pandoc, when available, to convert LaTeX content into readable Markdown or plain text. The original .tex file should remain preserved as the linked source, while the converted text is used for coding and analysis inside QualCoder.

The feature could remain optional so that QualCoder does not require Pandoc as a mandatory dependency. If Pandoc is not installed, QualCoder could display a clear message explaining that LaTeX import requires Pandoc.

I would be willing to work on the implementation and submit a pull request if this feature fits the project's scope.

ccbogel:
Hi, I hope you are enjoying QualCoder for your work.
What about using a python module such as:
pydetex 
pylatexenc

to store plaintext in the database

That way it would be an easy add to the _load_file_text()_ method

stoicfist:
Thanks for the quick feedback! I’ll test _pylatexenc_ and _pydetex_ and use whichever works best for the .tex import. That way, users can benefit from it without having to install Pandoc separately.

I’d be happy to work on this and submit a pull request.

ccbogel:
Thanks.
I have merged the pull request.
We intend to do a beta release in a few weeks. then a next release a month after that. 
Do you have a small sample .tex file, that could be added to the Examples folder?

stoicfist:
Yes, I can prepare a small self-contained example .tex file for the Examples folder. 
I will keep it generic and include a few common constructs such as _sections_, _emphasis_, _a list_, and a _simple table_, without _external images_ or _included files_.

ccbogel:
Sure, that would be lovely.

ccbogel:
Hello, yes that's fine. We can also move the conversation to ccbogel@hotmail.com

ccbogel:
Excellent

PR #1456, opened 2026-08-04
Add LaTeX file import

stoicfist:
## Summary

Adds direct `.tex` file import to QualCoder using `pylatexenc`.

The importer converts LaTeX source files into readable plain text for coding and analysis while preserving the original `.tex` file as a linked source.

## Changes

- Add `.tex` support to the existing Manage Files import workflow
- Convert LaTeX content to readable plain text with `pylatexenc`
- Preserve the original `.tex` source path
- Reuse shared text encoding detection
- Support replacing existing text sources with `.tex` files
- Add focused tests for LaTeX conversion and import behaviour

## Limitations

- `\input` and `\include` files are not resolved
- LaTeX is not compiled or executed
- Complex presentation-only constructs such as TikZ may not always convert perfectly
- Pandoc is not required or used

## Testing

Tested inside a Debian 12 Distrobox running on a Gentoo Linux host, using Python 3.11 and libVLC 3. 
_The implementation itself uses only Python and pylatexenc and does not depend on Distrobox or Linux-specific tools._

```bash
PYTHONPATH=src python -m unittest tests.test_latex_import
python -m compileall src/qualcoder/latex_import.py src/qualcoder/text_decoding.py
git diff --check
```

Refs ccbogel/QualCoder#1417

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,7 @@ json_repair
 mcp>=1.0.0
 markdown-it-py
 pyinstaller<=6.15 
+pylatexenc>=2.10,<3
 # The bootloader in the most recent version of pyinstaller is often falsely detected as malware
 # by anti virus software, so, it's safer to use a version that has been released a few months ago.
 pefile<2024.8.26  # see: https://github.com/pyinstaller/pyinstaller/issues/8832#issuecomment-2413203254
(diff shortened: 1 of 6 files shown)


ccbogel:
Thank you. good to see some other fixes and the separate module for text_decoding

PR #1460, opened 2026-08-05
Add sample LaTeX import file

stoicfist:
Add a small self-contained interview example for testing LaTeX document import.

Refs #1417


ccbogel:
Thank you

PR #1464, opened 2026-08-06
Memo Updates (for review)

Salomon-Mazatlan:
Memo editor updates:

Buttons were added to the memo editor to insert information and export its content.

image

- **Insert date**: inserts the current date and time on its own line, in bold.
- **Insert coded segments**: only segments belonging to the memo's entity can be inserted, not those of unrelated codes or files.
  - Case memo: only the codings found in the files that belong to the case.
  - File memo: only the codings within that file.
  - Category, code and sub-code memos: only the segments coded with them, including their descendants (sub-categories and sub-codes).
  - Project memo: all coded segments in the project.
  - Each insertion follows a structured pattern: quoted **CODED SEGMENT** / **CODED MEMO** plus a detail line with position, hierarchy path (Category > Code > Sub-code), case, file and coder, adding only what exists. Covers text, image and audio/video codings ([Image x y w h] / [A/V hh.mm.ss]).
  - The selection dialog includes a segment-type filter (which also clarifies whether the last column shows coded text or a memo), text search, multi-selection and double-click to insert.
- **Insert references**: same scoping logic, displayed in APA format with their linked files; a dropdown allows switching to "All references" to cross-reference across the whole project (if the entity has no linked reference, the dialog opens there directly).
- **Export memo**: exports as an .odt file, converting markdown conventions (headings, bold, italics) into real document styles. Inserted labels use bold markdown, already rendered by the editor's highlighter.
- The buttons only appear on memos with an entity; in other uses of the editor (date selector, graph text editing) the toolbar is not shown.

___

**Possible improvements:**

- Optionally allow inserting coded segments outside the memo's scope (as the "All references" dropdown already allows for references), e.g. with an "All segments" option in the selection dialog, to compare or link evidence across different cases, files or codes.


Salomon-Mazatlan:
> Note: this is a version I developed before the Memo manager proposed in the roadmap. It can serve as an entry point: it already covers inserting dates, coded segments and references into memos, and lays the groundwork for later features such as hyperlinks.

qwertygc:
Very nice

Issue #1471, opened 2026-08-11
Translated combobox options break exports


Salomon-Mazatlan:
**Bug**
Several comboboxes compare the visible text (`currentText()`) against hard-coded English literals. The .ui item labels are translatable, so in any language where they differ from English the match fails silently: the user picks an export option and nothing happens.

**Reproduce:** run QualCoder in (e.g.)"Farsi", open Reports > Code retrieval, run a report and pick any export option. No file is produced. 

<img width="1174" height="689" alt="Image" src="https://github.com/user-attachments/assets/7a1791cd-2372-4a09-aa15-98e041a11c93" />

**Affected**
- report_codes.py: comboBox_export, comboBox_sort (broken); comboBox_memos, comboBox_matrix (fragile: mixed `_()` comparisons and magic index numbers)
- code_pdf.py: comboBox_exports (broken)
- select_items.py: group filter, added as `_("All")` but compared to `"All"` (broken)
- code_text.py, memo.py, view_charts.py: fragile (index-based, translated strings as logic values, untranslatable heatmap items)

**Fix**
Canonical key in each item's `userData`, dispatch on `currentData()`. Labels stay translatable, no .ui changes, independent of translation and item order. Applied consistently to all combos above; also fixes an unwrapped "Also coded memos" literal in the matrix code. In comboBox_sort, `cat_az`/`cat_za` are reserved for the Category sort branches not yet in the .ui.

ccbogel:
thank you

PR #1472, opened 2026-08-11
Fix: Translated combobox options broke exports 

Salomon-Mazatlan:
Translated combobox options broke exports and other combobox-driven actions: dispatch code compared the visible currentText() against English literals, while .ui item labels are translatable. Issue: https://github.com/ccbogel/QualCoder/issues/1471 

Fix: store a canonical key in each item's userData and dispatch on currentData(). Labels stay translatable, no .ui changes, independent of translation and item order. Applied consistently across 6 files.

cat_az/cat_za keys are reserved for the Category sort branches not yet in the .ui.

### DIFF
--- a/src/qualcoder/code_pdf.py
+++ b/src/qualcoder/code_pdf.py
@@ -2193,6 +2193,9 @@ class DialogCodePdf(QtWidgets.QWidget):
             pass
         
         try:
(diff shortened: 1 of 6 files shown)


ccbogel:
Gracias

PR #1481, opened 2026-08-14
Code PDF:  Rotate page &  emits

Salomon-Mazatlan:
- Rotate the current page 90° left/right (two new buttons and context menu entries). Single-page view mode only and view-only: the PDF file and coded coordinates are never modified. Changing page clears the text selection to avoid accidental codings.

- Missing project_data_changed emits: annotations (create/edit/delete), coding memos and "important" flags, file memo, imported PDF annotations to memo, and the restructure journal entry. On receive, the dialog refreshes annotations when they change in another dialog. ([COVERED:](https://docs.google.com/spreadsheets/d/1E2vV-m99yGPby_GfDtorR8Lpf0z3-njasCtd4cAMOi4/edit?usp=sharing))


ccbogel:
gracias

PR #1497, opened 2026-08-17
Adjustments for PDF import

Salomon-Mazatlan:
- pdf_utils.py: shared PDF helpers; manage_references no longer loads code_pdf.
- PDF import detects underlines: "PDF Underlines" category with Underline+colour codes (comments go to the segment memo).
- Colours via color_matcher(), with an achromatic guard and exact mapping of the Zotero/Acrobat palettes.
- Extractors hardened against encrypted or damaged PDFs.
- Code PDF refreshes its file list on Zotero/RIS imports or deletions from other dialogs (event receiver fix).
- Bug fix: deleting in Manage files now removes the physical file too (upstream path bug), no longer aborts on permission errors, with a guard against paths outside documents and safe vectorstore removal.


ccbogel:
thank you

PR #1501, opened 2026-08-18
Update code_text.py

Salomon-Mazatlan:
- annotate() deletes by anid instead of pos0; clearing an annotation's text no longer removes annotations from other files or coders.
- shift_code_positions passes new_pos0 + 1 to substr() (1-based); stored seltext is no longer shifted one character.
- edit_mode_on always reloads the full DB text (was: and condition that skipped the first chunk); saving no longer overwrites the document with the visible fragment and loses the rest of the text, and the editor text now matches stored positions exactly.
- edit_mode_on/edit_mode_off remember and restore the chunk window; after editing, the file returns to the partial view instead of staying fully loaded.
- edit_mode_off also updates file_['characters']; after editing, "next 50,000 characters" no longer skips the first chunk.
- prev_chars clamps start to 0 before the line-ending scan; navigating back reaches character 0 instead of hiding the beginning.
- update_positions rewritten as a left-to-right diff walk with interval-based insert/delete helpers; spans adjacent to the edit point are no longer shifted or shrunk (edit-save-edit-remove now restores exact positions), one-keystroke replacements are handled, and the dead branch plus the "annotations" table typo are gone.
- ed_update_codings/ed_update_casetext clamp spans ending at the text end instead of deleting them; ed_update_annotations reads the correct deletion marker (newpos0).
- Text search covers the full file (not only the loaded chunk); fixes the "QTextCursor::setPosition: Position out of range" console warning.
- Search navigation keeps chunked loading: switching files loads the self.files entry with its partial window, and a match outside the loaded chunk loads the chunk containing it.
- Software citation updated

### DIFF
--- a/src/qualcoder/code_text.py
+++ b/src/qualcoder/code_text.py
@@ -190,6 +190,7 @@ class DialogCodeText(QtWidgets.QWidget):
         self.edit_original_annotations = None
         self.edit_original_case_assignment = None
         self.edit_original_cutoff_datetime = None
+        self.edit_prev_window = None  # (start, end) of a partial chunk view, restored after edit mode
         
         # For Code Resize Handles Experimental
         self.active_handles = []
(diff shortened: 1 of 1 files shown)


ccbogel:
That is a good lot of fixes, thanks

Salomon-Mazatlan:
Thank you.

I have closed the PR. I will make some adjustments to remove the chunked file navigation options, which should resolve a good number of the previous errors.

PR #1503, opened 2026-08-19
Update report_cooccurrence.py

Salomon-Mazatlan:
- Matrix misaligned after transposing and reprocessing. Counts were placed using a name list that was not updated on reorder, so they landed on the wrong code pairs. Now placed by cid.

e.g.:

image

- Transpose with categories selected. Headers were re-sorted alphabetically while the matrices were only reversed, misaligning cells. Now only the order is reversed.

e.g.:

image

- Touching segments counted as overlap. Two adjacent codings (end of one = start of the other) sharing no character counted as a co-occurrence. Now classified as Proximity and not counted.

e.g.:

image

- Graph edge weights. add_edge overwrites the weight instead of adding it when a pair has values in both directions. Now the upper triangle is traversed summing both.
- External refresh lost the transposition. Changes from other dialogs reset the table to alphabetical order while transposed. The order is now preserved.
- Merge did not refresh. After merging codes, neither the table nor other dialogs were notified. Now it reprocesses and emits project_data_changed.
- Yellow highlight off by one. The +1 for the newline was added at the end of the span instead of the start; with an empty text_before it painted the previous line. Spans are now exact.
- "Hide blanks" out of sync. After transposing, the checkbox stayed checked while everything was visible. Now it is unchecked.
- Excel did not match the view. The full matrix was always exported in alphabetical order, ignoring hidden rows/columns and the current order. Now it exports exactly what is visible, with independent row and column headers.
- Minor. Excel column widths computed over the visible columns, and a redundant fid filter removed since the SQL already applies it.

### DIFF
--- a/src/qualcoder/report_cooccurrence.py
+++ b/src/qualcoder/report_cooccurrence.py
@@ -161,6 +161,8 @@ class DialogReportCooccurrence(QtWidgets.QDialog):
         self.codes, self.categories = fresh_codes, fresh_categories
         if self.code_selection_mode == "all":
             self.selected_codes = deepcopy(self.codes)
+            if self.transposed:  # Keep the reversed view across external refreshes
+                self.selected_codes.reverse()
             self.selected_categories_string = ""
             self._rebuild_selected_code_strings()
             return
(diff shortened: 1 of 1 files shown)


ccbogel:
Gracias

PR #1506, opened 2026-08-21
Cosas de video 7

Salomon-Mazatlan:
- VLC video no longer flashes or freezes at the top-left corner (native-window guards, HWND geometry sync, release() on backend switch).
- Audio files no longer show a black video frame or leave an orphan floating window.
- Keyframe measurement from a previous file can no longer apply to the current one.
- Annotation deletion by anid; no longer removes annotations from other files or coders.
- Opening a project closes the current one first; tabs no longer show the previous project's data nor can write its ids into the new database.


ccbogel:
Gracias

PR #1508, opened 2026-08-21
Update Speakers

Salomon-Mazatlan:
Final pending adjustments worked on with Van:

- Saved regex, button next to the custom pattern field (visible only with Custom checked); loads, edits (double click, with validation) and deletes the patterns stored in speaker_regex.json; "Use" copies the selected pattern to the field and re-scans; changes persist on close.
- ':'-chained codes: name1: name2: text creates one table row per code, all coding the same segment.
- Chain safeguards: strict Name: pattern (up to 3 word tokens glued to the colon), guards against bare times (10:30), URLs, over-long tokens, duplicated names, blanked time stamps in the gap; max 8 per chain.
- select_items.py: checkbox-selection sync.
______
- Minor: _save_regex_store() extracted so the completer and the manager share one writer.


Salomon-Mazatlan:
This is the last pending update on my side. I have found some minor details along the other py files, but they do not affect how the software behaves; I will keep testing them and we can handle them after the 4-beta release.

PR #1509, opened 2026-08-21
fix fr

qwertygc:
I have translated Lorenzo’s translations into French. I found it difficult to translate ‘inline codes’ accurately; I hope I have understood what they are.


Salomon-Mazatlan:
Hi Justin, 

In the Mark Speakers module, a speaker marker can carry extra codes written inside the marker itself, separated by ';'. For example:

> [Ana; woman; migrant] Well, I started caring for my mother in 2019...

Here "Ana" is the speaker name, and "woman" and "migrant" become codes applied to that turn. I called them "inline" because they are embedded in the same line as the speaker marker, rather than typed by the user in the interface.

Would "marker codes", "embedded codes" or "per-turn codes" be clearer?

May I ask you a favour again? If you could update the Spanish language files, I have no further changes pending on my side for QC.

qwertygc:
OK. Maybe "embedded codes" it's more accurate. I change the translation in French.

Maybe @ccbogel can said the better expression for that, this a native speaker in english.

I can update in spanish but you need to review, I don't speak spanish...

Salomon-Mazatlan:
> OK. Maybe "embedded codes" it's more accurate. I change the translation in French.
> 
> Maybe @ccbogel can said the better expression for that, this a native speaker in english.
> 
> I can update in spanish but you need to review, I don't speak spanish...

Thanks a lot. I would only need the files updated so they pick up the new translatable strings, I will take care of the translations themselves.

On my machine the update still fails and I haven't found the cause: running it leaves me with empty (0 kb) files.

ccbogel:
embedded codes seems good to me

PR #1516, opened 2026-08-25
Merged project Issues (for Test)

Salomon-Mazatlan:
merge_projects.py

- bug fixes: pdf_page was not carried over: PDF coded areas arrived as NULL and piled up on page 1.
- bug fixes: copy_source_files_into_destination: absolute path join, so no file was ever copied.
- bug fixes: case_text: remapped against newid instead of id, losing case-file links or pointing them at the wrong file.
- bug fixes: code_text.avid was not carried over: transcript codings lost their A/V segment.
- bug fixes: code_image and code_av not deduplicated: merging twice duplicated areas and segments.
- bug fixes: Orphan codings were inserted with cid = -1 or fid = -1.
- bug fixes: A .qda folder without a project: an empty data.qda was created in the user's folder, then a crash.
- bug fixes: Empty project table: TypeError when reading the database version.
- bug fixes: Source connection was never closed.
- bug fixes: Case attribute placeholders written with attr_type 'file' instead of 'case'.
- bug fixes: case_text inserted without owner, date or memo.
- bug fixes: Only sqlite3.Error was caught: any other failure left an open transaction and a modal dialog on screen, freezing the application.
- bug fixes: A failing AI vectorstore update, which runs after the commit, propagated and left the same modal stuck.
- bug fixes: Debug print in insert_categories.

Updates:
- robustness and performance
- The whole merge now runs in one transaction on its own connection: cancel or failure rolls back cleanly, and timers in other dialogs can no longer commit it half way.
- One commit() per row removed. The database phase went from 3.69 s to 0.02 s with 8300 codings, and stays at 0.34 s with 100 000.
- Document text is no longer held in memory for every file. Peak extra memory on a 342 MB project dropped from 400 MB to 132 MB.
- Id remapping moved from nested loops to dictionary lookups.
- Media copied in 8 MB chunks, so Cancel responds inside a large video and no partial file is left.

- DialogMergePreview: preview of what will be incorporated, with counts that match what actually gets inserted, warnings in red, and an option to save the report to journals.
- MergeProgress: progress dialog with Cancel during reading, preview and file copying.

special_functions.py

- projects_merged was overwritten: cancelling a second merge undid the tab refresh owed to the first.
- The Merge button was only re-enabled on cancel, not on failure.

tests file:

test_merge_projects.py: 

- 103 headless checks covering all of the above, verified on Windows with Python 3.14.

### DIFF
--- a/src/qualcoder/merge_projects.py
+++ b/src/qualcoder/merge_projects.py
@@ -24,12 +24,82 @@ https://qualcoder.org/
 import datetime
 import logging
 from pathlib import Path
+import re
 import shutil
 import sqlite3
+
+from PyQt6 import QtCore, QtWidgets
 
 from .helpers import Message
 
 logger = logging.getLogger(__name__)
+
+# Max item names listed per preview section. Counts are always complete
+MAX_LISTED_ITEMS = 200
+
+
+class DialogMergePreview(QtWidgets.QDialog):
+    """ Read only preview of what a merge would add, by section with counts and names.
+    Nothing is written yet, so Cancel leaves the project untouched.
+    """
+
+    def __init__(self, app, sections, path_s, parent=None):
+
+        super(DialogMergePreview, self).__init__(parent)
+        self.app = app
+        self.sections = sections
+        self.setWindowTitle(_("Merge preview"))
+        self.setWindowFlags(self.windowFlags() & ~QtCore.Qt.WindowType.WindowContextHelpButtonHint)
+        self.resize(760, 580)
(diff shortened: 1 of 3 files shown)


Salomon-Mazatlan:
Hi @ccbogel 

I have made some updates and fixes to the Merge projects module. It has been working well for me, especially with PDFs that have coded image areas, which is a current problem that is now fixed. Feel free to try it out if you like.

ccbogel:
gracias

PR #1520, opened 2026-08-26
fix fr and remove HTML on translation

qwertygc:
To temporarily resolve the accessibility issue (HTML in the text), I have removed the HTML from the French translation. I’ll need to look into how to handle this in the source files (ui) (see other PR)

I have also updated merging file translation


ccbogel:
OK.  I will also have a look at the ui files.

PR #1521, opened 2026-08-26
Clean html (for review)

Salomon-Mazatlan:
- .ui/.py files with the HTML markup removed.
- **ui_html_report.csv**: a table of every modified string, before/after.
- **strip_ui_html.py**: cleanup script, to run again whenever the files are edited in Qt Designer (which puts the HTML back). With --check it only reports.
- **compile_ui_Windows.bat**: batch converter from .ui to .py with pyuic6, with a prior backup and atomic writes.

@qwertygc  
Feel free to review it. They work correctly on my computer, but I have not run in-depth tests.


ccbogel:
I am not sure how to download this separately, so I will commit the request and then check it.

PR #1523, opened 2026-08-26
Update ui_dialog_settings.py

maninthemiddle01:
In Settings → AI, each AI profile shows an "access info" link below the profile
selector — for Blablador the API-access guide
https://sdlaml.pages.jsc.fz-juelich.de/ai/guides/blablador_api_access/, for OpenAI
https://platform.openai.com/api-keys, etc.

This restores interaction on that links, make it clickable and copyable. Tested on my machine.


Salomon-Mazatlan:
Thanks for the fix

kaixxx:
Good catch, thank you, @maninthemiddle01. 
I think this broke as a side effect of some recent changes to the settings dialog. 
@ccbogel: Since this is such a small change and directly related to the AI settings, I am going to merge this quickly, ok? I will do the same with the second PR related to this.

ccbogel:
yes sure

PR #1532, opened 2026-08-27
Update ai_llm.py add Ragarenn

qwertygc:
Ragarenn is an AI service managed by a French university (and available to all French universities). It is a GDPR-compliant solution. Technically, it is an open-source stack comprising Open Web UI, Ollama and vLLM. I’ve been using it for a while now, and it’s very good.

### DIFF
--- a/src/qualcoder/ai_llm.py
+++ b/src/qualcoder/ai_llm.py
@@ -383,6 +383,20 @@ reasoning_effort = default
 api_base = https://openrouter.ai/api/v1
 api_key = 
 
+[ai_model_Ragarenn]
+desc = The University of Rennes is testing generative AI ("RAGaRenn")
+	since March 2024 to co-create usage guidelines with its staff. Using a 
+	local infrastructure on Eskemm Data, it aims to assess resource needs,
+	carbon impact, and professional use cases. 
+access_info_url = https://ragarenn.eskemm-numerique.fr/sso/ch@t/app
+large_model = ilaas/gemma-4-31b
+large_model_context_window = 32000
+fast_model = ilaas/mistral-small-4-119b
+fast_model_context_window = 32000
+reasoning_effort = default
+api_base = https://ragarenn.eskemm-numerique.fr/sso/ch@t/api
+api_key = 
+
 [ai_model_Ollama local AI]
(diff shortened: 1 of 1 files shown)


kaixxx:
Nice that we now also have an option from France, thank you. We should add it to the "Academic AI Infrastructure" section on qualcoder.org: https://qualcoder.org/doc/en/2.3.-AI-Setup/#academic-ai-infrastructure 

When it comes to the settings dialog in the app, my policy has always been to only include services in the list of default AI providers if they are available worldwide. As far as I understand, this is not the case with RAGaRenn, right? However, I am contemplating making an exception here, if you believe this service has a substantial user base in France. 

But in any case, there might come a point in the future where I must remove this service again to prevent the default list from getting too long (we have 12 entries already). Users could still add it as a custom AI provider, following the instructions on the website.

qwertygc:
I think ‘blablator’ is specific to Germany, but I might be wrong. In fact, it’s ‘ollama’, and simply changing the URL in that section would have been more than enough.

I’m running a Qualcoder training course at the start of the new academic year, the idea being not to discourage the use of this feature.

kaixxx:
In fact, "Blablador" is available worldwide (for academic use). Very generous. 

But let's include RAGaRenn for now and see if it works for people in France. I will, however, add a "(France only") note to the profile. 
You have specified gemma-4-31b as the main model. That works OK with QualCoder. But the new `Gwen 3.8 27b` is better and about the same size. I would switch if it becomes available on RAGaRenn.

qwertygc:
> In fact, "Blablador" is available worldwide (for academic use). Very generous.

Oh, it's amazing. I update my course about it.

I have this model
image

kaixxx:
> I have this model

I would stick with gemma-4 for now. But if they update Qwen 3.6 to 3.8, I would switch. Qwen 3.8 27b is all the rage right now. It comes close to the best models from OpenAI and Anthropic in many benchmarks. And it works very well with the new agentic AI in QualCoder, as it is optimized for such applications.

qwertygc:
Interesting. It would be good if you could post your benchmarks for the different models somewhere, especially for those who want to run an Ollama instance locally, so they know where to start.

PR #1533, opened 2026-08-28
Update manage_files.py

Salomon-Mazatlan:
Fix: When creating a new file, the code tried to disable rich text on a text box that no longer exists in that dialog (it is now a plain-text-only field, which never accepts formatting). That leftover line was removed.


ccbogel:
Gracias

PR #1539, opened 2026-08-31
Issues: AI Assistance (for review)

Salomon-Mazatlan:
Hi @kaixxx 

I found some issues when using AI Assistance in Code Text. The first is that the coding margin keeps showing the codings of the document that was in the viewer, and the segments found by the search appear as if they were coded, tooltips included. That can be confusing and can lead to deleting codings by accident.

There is also a problem when switching coding windows while the assistant is running.

On top of that, when the AI Agent assigns codes it leaves no trace in the memos, unlike what happens with AI Assistance. I added that trace and, along with it, the model being used, for traceability and auditing, and to keep both ways of coding with AI consistent.

This last part is a proposal, since academic journals increasingly ask authors to state the model used, not only the tool.

**Closed window fix**

- No more crash when the coding window is closed while an AI search is running.

```
  File "D:\QualCoder-master\src\qualcoder\code_text.py", line 6310, in <lambda>
    lambda chunks, session_id=current_session_id: self.ai_search_prepare_analysis(chunks, session_id),
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "D:\QualCoder-master\src\qualcoder\code_text.py", line 6380, in ai_search_prepare_analysis
    self.ui.plainTextEdit.setPlainText(

RuntimeError: wrapped C/C++ object of type QPlainTextEdit has been deleted
```

**Coding margin fix**

- Codes no longer float in the margin during a search, and the highlight colours no longer come back if something refreshes the project in the background.

image

- While the message is on screen there are no code tooltips and no margin clicks, and that text cannot be coded by mistake.
- Everything returns to normal once a result is opened or the file changes.

**AI model**

- The chat shows the actual model under the heading of each answer, not just the profile name, and stores it with the message so an old chat still says what produced it.
- The model is fixed when the answer starts, so it cannot change mid-run.
- What the agent codes carries the model in its memo, keeps it when the agent edits its own work later, and never stamps a human's memo.
- In AI-assisted coding the memo now separates profile and model.
- If the model cannot be determined, everything behaves as before.

image

### DIFF
--- a/src/qualcoder/ai_chat.py
+++ b/src/qualcoder/ai_chat.py
@@ -562,6 +562,7 @@ class DialogAIChat(QtWidgets.QDialog):
     current_streaming_chat_idx = -1
     chat_msg_list = [] 
     is_updating_chat_window = False
+    _chat_ai_model_snapshots = {}
     ai_semantic_search_chunks = []
     last_export_dir = ''
     # filenames = []
--- a/src/qualcoder/ai_chat.py
+++ b/src/qualcoder/ai_chat.py
@@ -2285,8 +2285,9 @@ class DialogAIChat(QtWidgets.QDialog):
             return heading
         safe_model = html_lib.escape(model_line)
         model_size = max(1, self.app.settings['fontsize'] - 1)
+        # plain label, matches the API field name and needs no translation
         heading += (f'<div style="font-size: {model_size}pt; opacity: 0.75;">'
-                    f'{_("model")}: {safe_model}</div>')
+                    f'model: {safe_model}</div>')
         return heading
 
     def _ai_agent_heading_html(self, author: str = '', chat_idx: Optional[int] = None,
--- a/src/qualcoder/ai_chat.py
+++ b/src/qualcoder/ai_chat.py
@@ -676,6 +676,7 @@ class DialogAIChat(QtWidgets.QDialog):
         self.ai_text_text = ''
         self.ai_text_start_pos = -1
         self._chat_ai_profile_snapshots: Dict[int, str] = {}
+        self._chat_ai_model_snapshots: Dict[int, str] = {}
         self.ai_output_autoscroll = True
         self._chat_window_refresh_pending = False
         self._chat_window_refresh_timer = QtCore.QTimer(self)
@@ -1702,6 +1703,7 @@ class DialogAIChat(QtWidgets.QDialog):
                                 msg_type TEXT,
                                 msg_author TEXT,
                                 msg_content TEXT,
+                                msg_model TEXT,
                                 FOREIGN KEY (chat_id) REFERENCES chats(id))''')
         
         cursor.execute('''CREATE TABLE IF NOT EXISTS topic_chat_embeddings (
@@ -1715,6 +1717,10 @@ class DialogAIChat(QtWidgets.QDialog):
         chat_columns = {str(row[1]).strip() for row in cursor.fetchall() if isinstance(row, tuple) and len(row) > 1}
         if 'compaction_frontier_msg_id' not in chat_columns:
             cursor.execute("ALTER TABLE chats ADD COLUMN compaction_frontier_msg_id INTEGER DEFAULT 0")
+        cursor.execute("PRAGMA table_info(chat_messages)")
+        msg_columns = {str(row[1]).strip() for row in cursor.fetchall() if isinstance(row, tuple) and len(row) > 1}
+        if 'msg_model' not in msg_columns:
+            cursor.execute("ALTER TABLE chat_messages ADD COLUMN msg_model TEXT")
         self.chat_history_conn.commit()
         self.current_chat_idx = -1
         self.fill_chat_list()
@@ -1760,6 +1766,7 @@ class DialogAIChat(QtWidgets.QDialog):
         self.ai_text_text = ''
         self.ai_text_start_pos = -1
         self._chat_ai_profile_snapshots.clear()
+        self._chat_ai_model_snapshots.clear()
         self._multi_chat_selection_active = False
 
         selection_model = self.ui.treeView_chat_list.selectionModel()
@@ -2204,6 +2211,13 @@ class DialogAIChat(QtWidgets.QDialog):
             author = ''
         return author or 'unknown'
 
+    def _current_ai_model_id(self) -> str:
+        """Model id actually used for requests, not the profile label."""
+        try:
+            return str(self.app.ai.get_active_model_id()).strip()
+        except Exception:
+            return ''
+
     def _normalize_ai_profile_author(self, author: str = '', chat_idx: Optional[int] = None,
                                      fallback_to_current: bool = False) -> str:
         """Resolve display/storage author names for AI messages."""
@@ -2231,7 +2245,26 @@ class DialogAIChat(QtWidgets.QDialog):
         author = self._current_ai_profile_name()
         if chat_idx is not None and chat_idx >= 0:
             self._chat_ai_profile_snapshots[int(chat_idx)] = author
+            self._chat_ai_model_snapshots[int(chat_idx)] = self._current_ai_model_id()
         return author
+
+    def _resolve_message_model_id(self, model_id: str = '', chat_idx: Optional[int] = None,
+                                  fallback_to_current: bool = False) -> str:
+        """Model id for one message: stored value, chat snapshot, then current."""
+
+        stored = str(model_id if model_id is not None else '').strip()
+        if stored != '':
+            return stored
+        if not fallback_to_current:
+            # Stored messages without a model stay blank, they predate this field.
+            return ''
+        if chat_idx is None:
+            chat_idx = self.current_chat_idx
+        if chat_idx is not None and chat_idx >= 0:
+            snapshot = str(self._chat_ai_model_snapshots.get(int(chat_idx), '')).strip()
+            if snapshot != '':
+                return snapshot
+        return self._current_ai_model_id()
 
     def _clear_chat_ai_profile_snapshot(self, chat_idx: Optional[int] = None) -> None:
         """Remove a previously frozen AI profile snapshot for one chat."""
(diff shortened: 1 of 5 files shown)


kaixxx:
Hi Lorenzo

Thank you! 
Some comments: 

> the coding margin keeps showing the codings of the document that was in the viewer and the segments found by the search appear as if they were coded, tooltips included. That can be confusing and can lead to deleting codings by accident.

Hm, I cannot reproduce this on my system. But your screenshot under "Coding margin fix" shows a real issue. If you've fixed it, thats fine.  

> There is also a problem when switching coding windows while the assistant is running.

Do you mean the crash when closing the window while a search is running? Yes, that seems to be a bug. Again, thanks for fixing.

> On top of that, when the AI Agent assigns codes it leaves no trace in the memos

Every addition by AI (code or case creation, codings, annotations...) is tracked by setting the "owner" field (= coder name) to "AI Agent". This is also shown in tooltips etc.:

grafik

Also, the "visible coder" filter I have introduced in January allows hiding all AI contributions if you want. This was the main reason to develop this mechanism - I wanted to make AI contributions traceable and controllable. I think this is a better mechanism than using the memo for this. 
In general, I'm a bit reluctant to add random technical notes to the memo, because in some methods (content analysis, grounded theory), such memos have a clearly defined methodological purpose. I think we should not mess with these notes unless the user clearly asks for it. 

> unlike what happens with AI Assistance. 

If you use the AI assisted coding feature, you are asked every single time if you want to add the AI interpretation to the memo. Thats completely optional. Also, it is not meant to track AI use, but more as as way of saving the sometimes quite detailed AI interpretations for later. In fact, codings done with the "AI assisted" function get the users name in the owner field, because they make the final decision if they want to code a result or not. 

> The chat shows the actual model under the heading of each answer, not just the profile name,

I understand your reasoning. Actually, I had this implemented at some point in the past, but removed it, because it often shows the same information twice: The AI profile is named "Anthropic Claude Opus 5", and the model "claude-opus-5". I think that would confuse people. 

However, adding the AI profile name to the interpretation memo in AI assisted coding is a good idea. Again, the model name is not really needed here in my opinion, because it is usually already mentioned in the profile name anyway. 

-----------------

So, in short: 
- I would keep the bug fixes and the addition of the AI profile to the interpretation memo in the AI assisted coding
- I would rather not keep your additions to the memos for AI Agent actions (coding), because they are tracked with the "owner" field instead.
- I would probably also not keep the additional model name in the chat, but that's debatable.

Salomon-Mazatlan:
Hi @kaixxx 

> I would keep the bug fixes and the addition of the AI profile to the interpretation memo in the AI assisted coding

Agreed. I am in hospital at the moment, so the updated PR will take a little longer. I will try to make the change tomorrow.

> I would rather not keep your additions to the memos for AI Agent actions

You are right that memos have their own methodological purpose in content analysis or grounded theory, and that technical notes should not be mixed into them. I had considered delimiting them with ##### so they would stay out of later analyses, but the underlying problem would remain, filling a space that is not meant for it with technical information. I will drop that part.

I agree that the "owner" field and the visible coder filter cover traceability for what the agent does.

> I would probably also not keep the additional model name in the chat

When the profile is called "Anthropic Claude Opus 5" you are completely right, that line is redundant. With aggregator providers it is not. With a single OpenRouter profile you can switch models as often as you like without touching the profile name, and that name identifies the service provider, not the model that answered. Something similar happens with Blablador, where one profile gives access to several models (I use this example because it is the one I have available):

image


image

with model:

image

There is a related point that I find more important than the chat line. In reports, when something is signed by the agent, all you get is that the AI Agent was the coder. The profile is not shown, and nothing else is either, so the report, which is exactly what gets shared and what reaches a reviewer, does not let anyone reconstruct or audit what the work was done with.

_____

**A different proposal, and I think a better one than the memo**: a new column in the database to record the model on the items the AI touches, and the option to break it down in reports only when asked for. That keeps the technical information out of the memos, does not bother anyone who does not need it, and still has it available for those who have to report it.

On where this came from, it was not a user request, but a literature review and the publishing guidelines of some journals, which are starting to ask authors to state the model and not only the tool.  For example: 

- Jones, K. M. L. (2025). Generative AI in Qualitative Research and Related Transparency Problems: A Novel Heuristic for Disclosing Uses of AI. International Journal of Qualitative Methods, 24. https://doi.org/10.1177/16094069251404329

- Meyer, F., Miggelbrink, J., & Nguyễn, P. (2026). Acceleration and Recursion in AI-Mediated Qualitative Research: Discussing Adapted Criteria for Qualitative Inquiry in the Techno-Epistemic Age. International Journal of Qualitative Methods, 25. https://doi.org/10.1177/16094069261459023

- Fehring L, Frings J, Rust P, Kempny C, Thürmann P, Meister S
Extension of the Consolidated Criteria for Reporting Qualitative Research Guideline to Large Language Models (COREQ+LLM): Protocol for a Multiphase Study. JMIR Res Protoc 2025;14:e78682. URL: https://www.researchprotocols.org/2025/1/e78682 . DOI: 10.2196/78682

- Qualitative Psychology (APA), AI policy. Specific to the use of generative AI in the qualitative research process. https://www.apa.org/pubs/journals/qua/ai-policy

Reporting guidelines from other fields that already suggest stating the model name and version:

- Lin, Z. (2026). A validity-guided workflow for robust large language model research in psychology. Behavior Research Methods, 58, Article 216. https://doi.org/10.3758/s13428-026-03073-2

- Gallifant, J., Afshar, M., Ameen, S. et al. The TRIPOD-LLM reporting guideline for studies using large language models. Nat Med 31, 60–69 (2025). https://doi.org/10.1038/s41591-024-03425-5

 -- https://www.tripod-statement.org/

Salomon-Mazatlan:
> Agreed. I am in hospital at the moment, so the updated PR will take a little longer. I will try to make the change tomorrow.

Or I can prepare the PR after the beta release.

qwertygc:
Take your time. Get well soon

kaixxx:
Hi Lorenzo

Sorry to hear that you are in hospital, I whish you all the best! 
Your health comes first. No need to update this PR. I can pick the bug fixes from it, and the rest is something we can discuss for the next release (4.0 final or 4.1). 

You have a lot of good ideas here. I agree that we might need new database fields to document more closely who did what in the process. The owner field was my first approach, but it's a little too simple. Not only does it not document the AI model used, it's also insufficient in situations where both humans and AI touched a certain code, coding or memo. We can discuss how to document this in more detail. For a publication, I think you don't need to lay this out for every single code. But it's a good information for researchers to enhance transparency and keep track of their own work. 
BTW, looking at your screenshots, you've found an elegant solution to integrate the model name in small print in the AI chat. That looks much better and is less distracting than my version that I've decided to discard. We can discuss if we incorporate that in the future.
BTW, when it comes to documenting AI use, QualCoder has massive advantages over almost all commercial packages: Not only the models used, but also the prompts are openly accessible, which enhances methodological transparency a lot. With commercial packages, you often don't even know the exact model it uses under the hood, and also the prompts are typically kept confidential.

Salomon-Mazatlan:
Hi @kaixxx , and thanks to @qwertygc too, for the good wishes.

On the point above: from what I have been able to document, and to test in other software (ATLAS.ti, MAXQDA and the QInsights platform), there is little transparency about what the AI actually does inside these programs.

In ATLAS.ti the only evidence left behind is the code group, and when you connect through an MCP server, what the AI did and what the user did become indistinguishable. I have just taught a course on exactly that, and from my point of view it is worrying. In MAXQDA there is a grouping plus a mention in the memo saying the content comes from the AI. In QInsights the human-AI interaction is built into S. Friese's methodology, although that interaction is heavily automated.

A summary of what I found, as an example:

image


QualCoder already has the best mechanism of the four, because the AI is a coder like any other and can even be compared with a human one

* Coder comparison with the AI Agent as one of the coders could be another proposal. I added it to the table because I enabled it in a modification of my own.

ccbogel:
Lots of good information here thank you all.
Yes, get well soon Lorenzo

Salomon-Mazatlan:
Hi @ccbogel, thanks a lot.

I will only work on the Spanish translation files, and I should have them finished during the day.

I have also added a few missing specifications to qualcoder4-0.md. What I am not sure about is how the note markers should be added there, or whether the notes are better written into the text without any markers.

kaixxx:
Hey Lorenzo,
That's a great comparison table. May I use it in my presentations? I would of course credit you as the source. 
Maybe the transparency regarding the underlying prompts of the AI features could be added to the table. In my experience, the prompt has an even bigger impact on the outcome than the underlying model. In QualCoder, you can not only inspect these prompts, but also alter them or add your own. As far as I can tell, this is not possible in any of the other packages (except for the very special case of using Atlas.ti via MCP).

Salomon-Mazatlan:
> Hey Lorenzo, That's a great comparison table. May I use it in my presentations? I would of course credit you as the source. Maybe the transparency regarding the underlying prompts of the AI features could be added to the table. In my experience, the prompt has an even bigger impact on the outcome than the underlying model. In QualCoder, you can not only inspect these prompts, but also alter them or add your own. As far as I can tell, this is not possible in any of the other packages (except for the very special case of using Atlas.ti via MCP).

Yes, it is work I am only just starting. NVivo and QDA Miner are still missing. QDA Miner lets you choose the provider through its API, at least it used to, and write your own instructions, so there is room to take the comparison further. I included QInsights mostly to have something to argue about with Susanne on LinkedIn :)

kaixxx:
I've talked to Normand Peladeau, the guy behind QDA Miner, on a mailing list. He is a positive example in advocating strongly for methodological transparency in AI use. On the QDA Miner website, he writes: "Prompt Transparency & Customization: View, edit, or create your own prompts. No black box" (https://provalisresearch.com/products/qualitative-data-analysis-software/qda-miner-features) (I haven't tested this directly in the software, though.)
Susanne also promised at some to publish the underlying system prompts of QInsights. But so far, I haven't found them. However, she is generally quite open in sharing prompting techniques that she uses in her analysis.

kaixxx:
Hi Lorenzo, 
I am extracting the two bug fixes from this PR and apply them. Everything else can stay for a later release. Thank you for pointing out the issues. Take your time to recover, we can work on this later.

kaixxx:
bug fixes included here: https://github.com/ccbogel/QualCoder/pull/1548 I've even managed to keep you as the commit author :)

Issue #1542, opened 2026-08-31
inside f-strings error (gettext)


qwertygc:
As discussed in our conversations, there are issues with line detection in gettext, which prevent the software from being translated properly. One solution is to change the way we translate.

> In the process, I‘ve found a problem in the translation files: Gettext “_()” calls nested inside f-strings are not extracted in the current POT files, which leads to nearly 200 strings not being translated properly.
> GNU gettext 0.23, released in December 2024, added Python f-string syntax recognition. But the current POT was probably generated with a pre-0.23 extractor.

 > I think as a general rule, we should avoid using the embedded _() translation method within f-strings.
> No: f"{_('My name is')}: {name}. "
> Yes: _("My name is") + f": {name}. "

Concerning files:
```
Detected embedded _() calls inside f-strings in src/qualcoder:
======================================================================

src/qualcoder/code_pdf.py
  Line 1906: tt += f"\n{_('Codings:')} {text_codings + area_codings}"
  Line 1908: tt += f"\n{_('Memo:')} {file_['memo']}"
  Line 3005: f" [{_('Page')} {pagina}: {ref['x1']},{ref['y1']} "
  Line 3868: text_ += f"[{_('Page')} {pagina}: {r[2]},{r[3]} {r[4]}x{r[5]}] " + _("Code: ") + f"{r[0]} ({r[7]})\n"
  Line 6206: cursor.insertText(f"{_('Project')}: {project_name}\n", header_fmt)
  Line 6207: cursor.insertText(f"{_('File')}: {self.file_['name']}\n", header_fmt)
  Line 6209: cursor.insertText(f"{_('Generated report')}: {report_date}\n\n", header_fmt)
  Line 6314: cursor.insertText(f"{_('Code')}: {c['name']}, {_('Coder')}: {c.get('owner', '')}\n\n", c_header_fmt)
  Line 6321: cursor.insertText(f"[{_('Co-occurring codes')}: {', '.join(sorted(co_codes))}]\n\n", norm_fmt)
  Line 6325: cursor.insertText(f"[{_('Coded memo')}: {coded_memo}]\n\n", norm_fmt)
  Line 6348: cursor.insertText(f"[{_('Page')} {page}] ", a_header_bold)
  Line 6349: cursor.insertText(f"{_('Code')}: {a.get('name', '')}, {_('Coder')}: {a.get('owner', '')}\n\n", a_header_fmt)
[...]

PR #1543, opened 2026-08-31
fix f-string error

qwertygc:
Fix the Fstring error (see #1542)


qwertygc:
I have tested, seem not error.

kaixxx:
As explained in my email (see below), I would not replace the f-strings in this way. This would most certainly create problems during translation.

My argument from the email:
> The problem is that the sentence structure is often different in other languages. So, `_("My name is") + f": {name}. "` may not translate to a correct sentence in any language. As an example, in Japanese, one would say “私の名前は{name}です。", which is roughly: “As for my name, {name} is.”
With f strings, you have more flexibility where to place “{name}” in the sentence, depending on the language. Why do you think using them is bad? An alternative would be the old `.format()` notation that I find a little cumbersome to read.

qwertygc:
I agree, with hindsight. I’ll have to work out how to incorporate this so that it’s manageable (particularly for my toolchain, which isn’t up to date). I can close the PR.

kaixxx:
No pressure. I can do the extraction for now. Then, we can find a solution that works for all of us for the next release.

qwertygc:
To be honest, I was a bit worried about getting the translation wrong (I make it a point of honour to localise software properly). But we can look at the bigger picture for version 4.1

PR #1544, opened 2026-09-02
Prepare v4 beta

kaixxx:
This is a collection of small fixes and updates I wanted to make for vers. 4.0

I have also updated all the language files. 

@qwertygc: In order to extract the language files, I had to update `rebuild_lang.py` to use Qt6 tools (pylupdate6). Before, it was still using the Qt5 tool set. Since we are using PyQt6 in QualCoder, I think it's safer to use Qt6 for the extraction as well. Note that pylupdate6 is no longer using project.pro, so I had to change some logic in the script as well. I hope that's OK for you. If not, feel free to go back to the old rebuild_lang.py

I still have to go over the German language files manually to check them; will create another PR ASAP.


ccbogel:
Thank you that is very extensive. So once your next PR comes in - will that then be the last thing before the 4.0-beta release?

ccbogel:
I downloaded the latest zip and ran that. 
A Message on opening - "A newer AI profile is available for ..."
As the Settings did not have enable AI integration checked. This message should not appear.

qwertygc:
OK, i need to update translation

kaixxx:
> As the Settings did not have enable AI integration checked. This message should not appear.

Yes, I'll fix that. 

From my side, this is everything I had on my list for v4.0 The only thing left is to check the German language translations. 
Once the French and Spanish translations are ready as well, I can create the binaries (Win/Mac), and we can release the beta by the end of this week, as planned.

qwertygc:
OK. Once the translations are ready, I create the linux binaries.

PR #1567, opened 2026-09-09
Toggle tooltips menu option - for QPlainTextEdit

ccbogel:


ccbogel:
Toggle tooltips menu option - for QPlainTextEdit

kaixxx:
May I suggest persisting this in config? Otherwise, it would not even survive a recreation of the coding dialog... Otherwise, a neat solution.

PR #1570, opened 2026-09-10
Sonal import

qwertygc:
[Sonal](https://sonal-info.com/)  is an open-source qualitative analysis programme that is particularly widely used in France. Whilst it is very powerful at linking sound to images (the use of waveform colouring is particularly advanced), making it well-suited for transcription, it has a number of limitations when it comes to analysis.
This pull request aims to enable the import of files from SONAL projects (which do not include REFIQDA) into QualCoder.

I drew inspiration from the Taguette file import structure to import files into Sonal.

We’ll have to see whether we integrate this into version 4.0 or 4.1.


ccbogel:
I have not heard of Sonal - it seems very niche software -this makes me unsure.
If you are sure it imports well, and that it is really widely used, then, OK we can add this now.

PR #1571, opened 2026-09-10
External MCP server access

kaixxx:
This PR adds an option to open the internal MCP server in QualCoder for external access from the local machine. 

**Background:** 
- MCP (model context protocol) is an open standard from Anthropic for AI Agent frameworks like Claude Code or OpenCode to interact with applications. It is used widely in the field. 
- QualCoder's internal AI agent is already based on this protocol. I did this to facilitate interoperability with external agent frameworks later on. However, for vers. 4.0 beta, I ran out of time. 
- Since then, two other QualCoder MCP implementations have popped up: qualcoder_mcp and qualcodernext. Both operate directly on the project database, bypassing QualCoder altogether. I think this is not a good approach, as it might lead to data corruption and also bypasses all the safeguards regarding AI use built into QualCoder. Granted, both frameworks come with their own safeguards. But I think it is a bad user experience if people set the AI permissions to "Read only" in QualCoder but an external tool like Claude Code still has full write access via direct database manipulation.
- This is why I think we should take the lead here and release an official QualCoder MCP server as soon as possible (with 4.0 final). This is not very complicated, as the server itself is already in place and tested. This PR mainly adds the communication interface for external frameworks. 

**Implementation and safeguards:**
- The external MCP access must be enabled via the settings dialog.
- It opens a small web server on localhost for MCP communication.
- It respects the AI permissions set in QualCoder (Read only / Sandboxed / Full access).
- All items created via MCP are clearly marked in the owner field with "External MCP". 
- MCP only works when QualCoder is running and can only access the current project in the app. 
- Changes made via MCP show up "live" in the UI (e.g., the code tree). This makes controlling the operations of an AI agent much easier. 
- All changes made via MCP are recorded in the Undo trace for AI actions (AI Agent window) and can easily be reverted.


kaixxx:
Small addition: This PR also updates the MCP SDK to vers. 2 and adds uvicorn, a very lightweight web server, to the requirements.

ccbogel:
This has introduced an error:
No module mcp.server.context     Issue #1572

PR #1581, opened 2026-09-14
Import sonal pi amelior

PUD-GA:
- the `.crp` file is used as the metadata source of truth;
- SonalPi's speaker convention is respected: names ending in `?` are interviewers, while the other speakers are respondents;
- speaker-scoped variables are aggregated into one QualCoder file attribute without allowing an interviewer's empty value to hide a respondent's value;
- speaker names are added to the imported transcription while offsets are still calculated correctly;
- a parent theme remains a QualCoder category, but its twin code is only created when that parent is directly used in a transcription; leaf codes remain available;
- whitespace-only thematic spans left behind by SonalPi are discarded when isolated, but retained when they connect text-bearing spans carrying the same code;


ccbogel:
thank you

PR #1591, opened 2026-09-19
Delayed loading of AI components for improved startup time

kaixxx:
This implements the delayed loading of AI components, which will improve startup time. If AI is disabled, AI components are not loaded at all.

The loading is divided into two steps: 
1) vectorstore and embedding components (the "AI memory" or search index)
2) connection components like langchain, etc.

The vectorstore is also loaded if only the external MCP server is active. This way, external MCP becomes usable independent from the internal AI setup. 

(As a side note: This delayed loading should also allow to compile a strict "no AI" version were activating AI integration is technically impossible, if this request comes up again in the future. By leaving out langchain, openAI, sentence-encoder, and other AI packages in a pyinstaller compile, QualCoder should still run like normal with disabled AI, but activating it would fail. This is untested, though.)


qwertygc:
Nice thanks, and good idea the no ia version

kaixxx:
> good idea the no ia version

Haha, just to clarify: I am not advocating such a crippled version. I think people should have the choice, and AI is disabled by default anyway. However, in the past, we have had requests from an IT department with such a strict anti-AI policy that they would not allow any software where AI integration was technically possible. I just wanted to present an option in case such requests come up again.

ccbogel:
It's good to have a no AI option. But mainly good to have a faster startup. regards Colin

qwertygc:
> > good idea the no ia version
> 
> Haha, just to clarify: I am not advocating such a crippled version. I think people should have the choice, and AI is disabled by default anyway. However, in the past, we have had requests from an IT department with such a strict anti-AI policy that they would not allow any software where AI integration was technically possible. I just wanted to present an option in case such requests come up again.

I'm having discussions during training sessions where people have concerns about the GDPR (and privacy officers are rather apprehensive). Having an option to reassure them would be helpful.

